PR #50, opened 2026-08-16
feat(integrations): cloud controller TLS + dedicated AWS, Azure and GKE screens

belliel:
## What changed

- **AWS, Azure and GKE each have a dedicated integration screen** — node pools, machine types, spot status, and now the ingress picture specific to each controller (ALB groups, AGIC annotations, GKE managed certificates).
- **`TlsBadge` reads vendor-managed TLS** — all three managed controllers keep the certificate off `spec.tls`; the Ingresses column now shows what the controller terminates rather than reading "no TLS" on every HTTPS site in a managed cluster. A tooltip names the source and the covered hosts.
- **`IngressList` fetches TLS once per page**, passed to cells through a `VendorTls` context so a list of fifty ingresses is still one request.
- **`ResourceList` accepts an `error` prop** for pages that fetch their own rows — previously a failed read and an empty scope arrived identical and the table said "no resources" for both.
- **Column and quick-action arrays are memoised** in `ResourceList` — rebuilding TanStack's column model on every two-second watch tick was remounting action buttons mid-hover.
- **`PodList` passes a `headerContent` slot** to `ResourceList` instead of wrapping it in a `div`, so the metrics banner is inside the height-filling column rather than outside it.
- **`scripts/gen_icon.py` deleted** — it drew "K8" letters from before the product was called Rubick; running `make gen-icons` would have quietly replaced the real icon. `base.png` is now the committed source of truth.
- **Pre-push hook removed from `lefthook.yml`** — it ran the full frontend suite on every push, long enough to time out a tool driving git. CI already runs a strict superset; the tradeoff and what to do instead are documented in the file.

## Why

Rubick is a GUI for Kubernetes clusters, including managed ones on AWS, Azure and GKE. Those clouds keep TLS off the Ingress object the controller owns, so the app was telling users their HTTPS sites had no TLS. Giving each cloud its own screen also surfaces things the generic ingress list cannot — ALB sharing across namespaces via `group.name`, GKE `ManagedCertificate` status, Azure Workload Identity bindings.

Closes #

## Test plan

- [ ] Verified manually on a GKE cluster (the primary dev target)
- [ ] AWS and Azure paths covered with unit tests against the controllers' documented behaviour — issues against a live EKS/AKS cluster are very welcome
- [ ] `cargo fmt`, `cargo clippy`, and `bun run lint` pass
- [ ] New tests: `ResourceList.test.tsx`, `useResourceDetail.test.tsx`, `data-table.test.tsx`, `aws/groups.test.ts`, `aws/ingress-tls.test.ts`, `azure/ingress-tls.test.ts`, `azure/workload-identity.test.ts`, `google-cloud/routes.test.ts`, `cert-manager/serves.test.ts`, `loki/coverage.test.ts`, `prometheus/coverage.test.ts`, `traefik/model.test.ts`, `traefik/routes.test.ts`, `argocd/data.test.ts`


belliel:
Superseded by #51 — same commits, branch renamed from `t3code/cb799d51` to `t3code/cloud-controller-tls-and-screens`. A pull request cannot change its head branch, so this one is closed rather than moved.

PR #72, opened 2026-08-19
fix: short-lived certificates no longer read as almost expired from birth

### LINKED ISSUE #68: shortlived сертификаты всегда отображаются как почти просроченные
У letsencrypt есть shortlived профиль, который выдаёт сертификаты на 7 дней. Но во многих местах Rubick ожидает серты ну хотя бы на плюс/минус месяц, и начинает выдавать различного рода варнинги сразу после выдачи сертификата. Хотелось бы, чтобы длительность сертификата учитывалась в такого рода диагностиках.

Image

Image

У Certificate есть поле renewalTime, возможно, стоит использовать его и до даты его наступления вообще никак не помечать серты как "близки к дате просрочки".

belliel:
## What is wrong today

Let's Encrypt's short-lived profile issues 7-day certificates. The app's expiry thresholds are absolute — warn inside 14 days, err inside 3 — so a short-lived certificate is *born* inside the warn window and wears a warning from issuance to renewal, on every surface, forever. The one line built to matter ("expires in 4 days") drowns in permanent noise. Closes #68.

Two more problems surfaced while fixing it: the sidebar said nothing about a vendor page in trouble until you opened it, and switching clusters kept the previous cluster's integration numbers on screen — the count and page queries were not keyed by context.

## What changes

- **Thresholds become caps, scaled by lifetime.** Warn at `min(14d, lifetime/3)`, err at `min(3d, lifetime/10)`. For a 7-day certificate the warn lands exactly on cert-manager's default renewal point; a 90-day certificate never notices the difference. This fixes every surface at once — Secret pages, ingress findings, traffic chains — because they all go through `expiryOf`.
- **Where the Certificate CR is at hand, the verdict trusts the plan.** Before `status.renewalTime` a would-be warn reads **"renews in N days"** with no tone; past it the row says **"renewal overdue — expires in …"** and the cluster facts say **"N renewals overdue"**. The err threshold stays a backstop no schedule argues with. (Rejected alternative: renewalTime alone, as the issue suggests — Secret-only surfaces never see the CR and would have kept crying wolf.)
- **A dot beside the sidebar count** when the vendor's page found trouble (warn/err), nothing when it is quiet — a generic `PageCount.tone` contract, cert-manager is the first to use it. A spinner sits beside the Integrations caption while the category is still being read.
- **Every integration read is keyed `[context, ...key]`** — the same rule the detection scan always stated — so cluster B never reads cluster A's cache, and pages keep sharing the sidebar's cache entry.

## How to verify

`bun run test` — 1863 tests, including new ones pinning each state (scaled thresholds, renews-in/overdue verdicts, facts copy, sidebar dot, cache scoping across a context switch).

To see it live without waiting for a real short-lived issuance: create a kwok cluster, apply the cert-manager CRDs only, create Certificates and write their statuses by hand — `kubectl patch certificate X --subresource=status --type=merge -p '{"status":{...}}'` with `notBefore`/`notAfter` seven days apart and `renewalTime` in the future (quiet, "renews in N hours") or the past ("renewal overdue").

## Risk and rollback

Presentation only — no data, no migrations; revert restores the old marks. The one behavioural edge: certificates with an unreadable `notBefore` keep the absolute 14/3 thresholds.


Dudude-bit:
Reviewed, and pushed the four follow-ups onto this branch rather than leaving them as comments. The approach itself I would not change: scaling the thresholds by lifetime fixes every surface through `expiryOf`, and keeping `managedExpiryOf` separate is what lets the Secret-only surfaces stay honest — which is why "renewalTime alone", as the issue suggested, would have been worse.

**`babdfc7` — the cluster facts were not keyed by cluster.** The description says *every* integration read is keyed `[context, ...key]`. The counts are; `["integration-facts", vendor.id]` in `src/integrations/index.ts` was not, with a 60s `staleTime` and no invalidation on switch. Measured before the fix:

```
prod-eu facts : ["1 certificate","1 renewal overdue","Show it"]
staging facts : ["1 certificate","1 renewal overdue","Show it"]
listCustomResources calls: 1
```

The second cluster is never asked. It predates this PR — `main` does the same — but this PR is what makes facts loud: a stale count is wrong, "1 renewal overdue" pointed at the wrong cluster is an accusation. `src/integrations/facts.test.tsx` pins it and fails on the old key.

**`731f275` — three things the hours step left half-done.**

- Every caller still ranked by whole `days`, so certificates expiring inside one day tied and fell back to `localeCompare` — the alphabet deciding order in the hour order is the only thing the page is for. `Expiry.left` carries the exact remainder; the table, the facts and the ingress detail rank by it.
- `N expiring soon` lost the one thing the line was for. It reads `2 expiring, soonest in 6 hours` now, from the certificate the sort just proved is soonest.
- `span()` under an hour returned `expires in 0 hours` — the rendering-bug sentence the hours step exists to avoid, one unit down, in the last hour of a certificate's life. Minutes now, floored at one.

Each is pinned by a test I checked discriminates: reverting the change turns it red.

`bun run test` 1869 passed (three consecutive runs — two fixtures I first wrote sat exactly on a rounding boundary and flaked, moved off it) · `bunx tsc --noEmit` clean · `bun run lint` zero warnings.

One thing I did not touch: `CONTRIBUTING.md` asks for one topic per PR, and this is three. The commits are cleanly split, so it reads fine — but `fix(integrations): key every integration read to its cluster` is the one that touches every vendor at once, and its blast radius is not the same as a copy change. Worth its own PR next time.

Issue #70, opened 2026-08-19
Build Format: AUR / Arch Linux / pkg.tar..zst


syorito-hatsuki:
Extend CI with AUR package format or with just pre-builded `pkg.tar..zst` suitable for as `-bin` install on Arch.

https://tauri.app/distribute/aur/ - Maybe not the best way, but at least from official docs

Prototik:
I already maintain a [source package](https://aur.archlinux.org/packages/rubick-kubernetes). There is no problem for me to package a -bin version as well.

syorito-hatsuki:
> I already maintain a [source package](https://aur.archlinux.org/packages/rubick-kubernetes). There is no problem for me to package a -bin version as well.

For me it's problem to use source, cause R7 2700 takes around half hour to build Rubick😅

belliel:
The release CI already publishes what a `-bin` package needs: every release ships `Rubick_<version>_amd64.deb` (and an `.rpm`) — see the v4.3.0 assets. That's the same artifact the official Tauri AUR guide's `-bin` PKGBUILD consumes.

@Prototik if that works for you, a `rubick-kubernetes-bin` based on the deb would be very welcome — we'll keep the asset naming stable on our side.

Prototik:
[Done](https://aur.archlinux.org/packages/rubick-kubernetes-bin)

belliel:
That was fast — thanks @Prototik! 🚀

@syorito-hatsuki [rubick-kubernetes-bin](https://aur.archlinux.org/packages/rubick-kubernetes-bin) is up, no more half-hour builds. Closing as resolved.

PR #73, opened 2026-08-19
docs: record which release files the AUR -bin package depends on

belliel:
## What's wrong now

Since #70, the AUR package [rubick-kubernetes-bin](https://aur.archlinux.org/packages/rubick-kubernetes-bin) (maintained by @Prototik, outside this repo) hardcodes two things from us: the asset URL `releases/download/v${pkgver}/Rubick_${pkgver}_amd64.deb`, and raw pinned-commit URLs to five files under `src-tauri/icons/`. Nothing in the repo says so — a product rename, a bundle-target change, or an icon shuffle would pass review here and break the package on its next update.

## What changes

Docs only: a comment on the `linux-x64` matrix entry in `release.yml` naming the asset contract, and a new `src-tauri/icons/README.md` listing the externally fetched icons.

## How to verify

Compare both notes against the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rubick-kubernetes-bin): `source_x86_64` is the deb URL, `source` lists the icons.

## Risk / rollback

None — a YAML comment and a README. The bundler references icons by explicit paths, so the new file in that directory is inert.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # The AUR package rubick-kubernetes-bin
+          # (https://aur.archlinux.org/packages/rubick-kubernetes-bin)
+          # downloads releases/download/v${pkgver}/Rubick_${pkgver}_amd64.deb
+          # from every release. The .deb asset name is a public contract —
+          # renaming the product, the target, or the bundle format breaks
+          # the package.
           - name: linux-x64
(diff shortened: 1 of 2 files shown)


Prototik:
Just to clarify: that's usually not a big deal when upstream changes something in the distribution. Old links works fine (they pinned by commit), release assets kept intact too (like .deb). So feel free to change whatever you need - I'll catch it on package update anyway.

PR #76, opened 2026-08-20
Gateway API v2: full support with policy attachment and internationalization

### LINKED ISSUE #69: Поддержка Gateway API
Как альтернатива Ingress, который feature freeze.

belliel:
## What changed

Gateway API v2 support is now complete end-to-end:

- **Backend**: Added `BackendTLSPolicy` to the cross-cluster search. Routes now report their Gateway parents, attachment status, and conditions (`Accepted`, `ResolvedRefs`). Service connections show incoming traffic from Gateway API routes with the same stops and paths as Ingress.
- **Frontend**: Route detail pages trace traffic from listeners through routes to backends with live status. Route lists show hosts, backends, and gateways per row. Gateways list their listeners and attached routes; a listener port dials directly into the traffic chain. Peeks show namespace contents and Gateway specs with live policy attachment.
- **Internationalization**: Every Gateway API surface (detail pages, route lists, topology, trace steps, backend states, policy reasons) now speaks the cluster's language.
- **UI polish**: Probe commands for DNS and TCP from the client's network, clickable ports that forward to the traffic chain, hostnames and addresses one click from clipboard, policy attachment status in details and route conditions.

## Why

Gateway API 1.5+ is now stable across every major distribution. The v1 feature set in this app was incomplete (missing policy attachment, limited traffic visualization). This branch fills both.

Closes #69

## Test plan

- [x] Verified manually on macOS with live k8s cluster (GKE, kind)
- [x] Added integration tests: `live_gateway.rs` covers route detail chain, all five route kinds, Backend TLS policy reverse lookup, listener peek-through
- [x] Added unit tests: route trace, traffic doors, route rows, peer-sources, gateway topology
- [x] Frontends all render: GatewayClass, Gateway detail, route lists and details (all five kinds), policies
- [x] `cargo fmt`, `cargo clippy`, and `bun run lint` pass


Dudude-bit:
Merged, in 4.6.0 — through #84, which carried these 65 commits onto current `main` and added the fixes a review of them turned up.

The rebase was the substantial part: this branch predates main's i18n pass, so `describeStop`, `peek-sources` and `CrdColumn.header` came back on their pre-i18n signatures. Resolving that surfaced two real defects — `GatewayClass` was cluster-scoped in the frontend registry and namespaced in Rust, and the stand's five alpha kinds asked the API for `v1`, which they have never served, so they were silently never created.

Thanks — the shape of it is good, and the row list's three-state `verdictsKnown` is what the other surfaces were changed to copy.

PR #89, opened 2026-08-29
fix: read ExtensionRef-filtered rules with no backends as configuration

### LINKED ISSUE #69: Поддержка Gateway API

belliel:
## What changed

Routes whose rules carry an `extensionRefs` entry (e.g. Envoy Gateway's `HTTPRouteFilter` for direct responses) and no `backendRefs` are now treated as valid configuration — the same way redirect-only routes already are — rather than as breakage.

Concretely:

- Added `selfAnswered()` alongside `redirectOnly()` in `route-trace.ts`; it covers every rule that either redirects or defers to an extension filter, with no Service backends anywhere in the route.
- `backendSteps()` now branches on `selfAnswered()` and picks the right label (`gwRedirectsOnly` vs. `gwFilterAnswers`).
- The route board (`route-rows.ts`) shows the filter note in the row tail.
- `peek-sources.tsx` and `GatewayRouteDetail.tsx` surface the new label in the detail panel and rule rows.
- New i18n keys added in both `catalogue.ts` (EN) and `ru.ts` (RU).

## Why

An `ExtensionRef`-filtered rule with no backends is intentional — the filter handles the response itself. Flagging it as "no backendRefs — matched requests have nowhere to go" was a false alarm that made valid Envoy Gateway direct-response configs look broken.

Closes #

## Test plan

- [x] `route-trace.test.ts` — ExtensionRef-only rule reads as `serving: true` with `state: "ok"`; mixed redirect + ExtensionRef route also passes
- [x] `route-rows.test.ts` — ExtensionRef-only route lands in `serving` with "filter" in its tail
- [x] `cargo fmt`, `cargo clippy`, and `bun run lint` pass
- [ ] Verified manually on Linux (AppImage) against a cluster with an Envoy Gateway `HTTPRouteFilter` direct-response rule


src/lib/route-trace.ts
```diff
@@ -143,15 +143,32 @@ export const HOSTLESS_PROTO: Record<string, string> = {
...
+
+/** The wider no-backends configuration: every rule redirects or hands the
+ *  request to an ExtensionRef filter — Envoy Gateway's direct-response, for
+ *  one. What a vendor filter does is its vendor's business; that such a
+ *  rule needs no backend is the route's own word. */
+export function selfAnswered(route: RouteInfo): boolean {
+  return (
+    route.rules.length > 0 &&
+    route.rules.every(
+      (rule) => rule.hasRedirect || rule.extensionRefs.length > 0
```
Prototik, line 184: Do we really want to consider any extension as self-answered route? That's not the case for most of them, including Envoy Gateway's HTTPRouteFilter: it can contain only additional filters and rules and do not define a direct response, in that case it's not self-answered.

belliel: Checked against the spec — you're right that presence isn't the signal. The spec's own wording conditions on what the filter does ("if there are also no filters specified *that would result in a response being sent*, the implementation must actively reject…"), and EG's HTTPRouteFilter is indeed one kind for both directResponse and rewrites, so even a kind whitelist wouldn't hold — telling them apart means reading the vendor CRD's spec, which rubick doesn't do.

What I don't want to lose is the other half of the fix: reporting these rules as "a matched request has nowhere to go" is equally wrong — rubick can't claim breakage it hasn't verified either.

So I'll move this case to the trace's existing `blind` state instead of `ok`: the route stops being reported as broken, but the tool no longer asserts the filter answers — "an extension filter takes the rule; whether it answers without a backend is its vendor's contract, not judged here". Same mechanics as any other step rubick couldn't read: `serving` stays true, `servingKnown` goes false.

belliel: Done in e3b09a3: the backend steps go `blind` for a filter-only route (serving stays true, servingKnown goes false), redirect-only keeps `ok`, and the copy names the filter instead of vouching for it. Also tightened the guard from "no Service backendRefs" to "no backendRefs at all" — a route with a non-Service backendRef does name somewhere to go.

Prototik: Yep, I confirm all my routes now green and grey, no false red ones

PR #91, opened 2026-08-29
The drain waits

Dudude-bit:
Stacked on #90 — merge that first; this branch is #90 plus one commit.

## Why

#90 stopped the drain deleting past a PodDisruptionBudget. It also had to stop the dialog claiming the drain would wait, because it did not: one pass, then a list of what stayed. That was honest, but it left Rubick able to do less than `kubectl drain`, which asks again until the budget lets go.

This puts the waiting back, for real this time.

## What it does

A drain is now a long operation, shaped like the cross-cluster search that had already solved the same problem: `start_node_drain` returns a handle, progress arrives as `drain-progress` / `drain-finished`, and `cancel_node_drain` stops the asking. It re-lists the node and re-evicts until nothing is left that waiting could change.

Only a 429 is worth asking again — the other three refusals need an answer from the operator, so retrying them would be a loop over a question the cluster cannot answer. That is one line, `DrainRefusal::worth_asking_again`, and it has its own test.

The dialog stays open and says which try it is on, what has moved, and who is being waited for. "Still going" and "stuck" look identical on a spinner. It cannot be dismissed by a stray click while a drain runs, and leaving the window does not stop one — pods are already moving, and closing a window is not a decision to abandon a cluster operation half-done.

## Two things worth a reviewer's eye

**The event payloads are hand-mirrored.** The type generator only emits types a *command signature* reaches, and these live in `AppEvent` — the same reason `useResourceSearch` writes out `SearchHit` by hand. What that convention has never had is a way to notice drift, so `the_shapes_the_frontend_mirrors_by_hand` pins every field name and variant spelling on the Rust side and names the TS file to change. I would rather have had the generator do it; this is the honest second best.

**`failed` means two different things** and I got it wrong first — TypeScript caught it. The `failed` *phase* is the command being refused before any drain existed. The `failed` *outcome* is a drain that was running and broke, so it arrives with a report of what it had already moved. Both are handled, and the second has a test.

## Verification

- 19 Rust tests in `drain`, 16 dialog tests. Sabotage-checked: making the loop stop waiting fails one, dropping the finish event fails another.
- One test drives the whole task end to end against a port nothing listens on — gate, pass, error classification, finish event, session cleanup.
- `cargo clippy -D warnings` clean, 462 Rust tests, 2096 frontend tests, build green.


Dudude-bit:
Reopened as #92 — this one's base branch went with the #90 merge, which closes a PR rather than retargeting it. Same commit.

PR #111, opened 2026-09-04
fix: a client certificate returned by a credential plugin now reaches the cluster

belliel:
## What is wrong today

A credential plugin that answers with a client certificate instead of a token (Teleport's `tsh`, corporate PKI plugins) gets its certificate silently dropped. The plugin returns PEM in `clientCertificateData` and `clientKeyData`, as the ExecCredential API specifies, and the app copies the PEM straight into the kubeconfig fields `client-certificate-data` and `client-key-data`. kube reads those through a base64 decode, the decode fails, kube swallows the failure with `.ok()` and builds the client with no identity at all. The server then refuses an anonymous request, and the error names no cause.

## What changes

The PEM is base64-encoded before it is written into the two kubeconfig fields, which is what those fields hold in a kubeconfig on disk. The fix is Igor DC's commit from the `igordcard/rubick` fork, cherry-picked with authorship kept; a unit test on `apply_exec_credentials` is added on top, and fails with the PEM stored as it came.

## How to check it

`cargo test --lib -p k8s-gui auth::interactive`. On a cluster whose exec plugin returns a certificate, connect from the app: with this change the requests are made with the certificate and the lists load; without it every list shows the refusal.

## Risk and rollback

Only the two certificate fields of an exec answer change; token answers are untouched. Reverting the two commits restores the previous behaviour.

### DIFF
--- a/src-tauri/src/auth/interactive/mod.rs
+++ b/src-tauri/src/auth/interactive/mod.rs
@@ -139,14 +139,24 @@ fn apply_exec_credentials(
     auth_info: &mut AuthInfo,
     status: ExecCredentialStatus,
 ) -> Option<chrono::DateTime<chrono::Utc>> {
+    use base64::Engine;
+
     if let Some(token) = status.token {
         auth_info.token = Some(SecretString::from(token));
     }
+    // A plugin hands back PEM, but the fields it lands in are kubeconfig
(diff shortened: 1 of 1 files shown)


igordcard:
Thanks, you picked it up, no need to fork locally anymore :)

PR #122, opened 2026-09-06
fix: work the deferred audit findings to completion

Dudude-bit:
Works through the deferred audit findings to completion. Four were real and are fixed here; two belong to the held `deps/kube-4` branch because the types they need do not exist on `k8s-openapi` v1_31; three did not survive a look at the code.

## Fixed

- **A broken `config.toml` no longer kills the app.** The parse ran in Tauri's `setup` before a window existed, so an unreadable file left the app with no window and nothing said — the exact silent-failure class this project is about. It is now moved aside under a timestamped backup, defaults take over so the app opens, and the incident reaches the reader through the Diagnostics panel (`settings_recovered_finding`). The strict `load()` stays strict, so a save path still refuses to write over a file it could not read.
- **Config maps serialise in a stable order.** Ten `HashMap` fields became `BTreeMap`, so `config.toml` no longer reshuffles its sections on every write (Rust randomises `HashMap` iteration per process).
- **A `Forbidden` with no code is still a refusal.** `is_refusal()` asked only `code == 403`, but the 401 path already knows some distributions send the reason with the code unset. A 403 arriving as `reason == "Forbidden"` with `code == 0` was read as a hard failure and threw the reader out of a cluster their token could still use. Now it mirrors the 401 handling.
- **Dropped a dead dependency.** The direct `tokio-tungstenite` dep was unreachable: nothing produced its error, so its `From` impl and the `Error::WebSocket` variant it fed were dead. `kube` keeps its own copy for exec/attach.

Every fix has a unit test, each sabotage-checked.

## Deferred to `deps/kube-4` (cannot compile on main)

- **Pod-level resources (KEP-2837).** `PodSpec.resources` does not exist in `k8s-openapi` v1_31 — only from v1_32. The two aggregation sites (`pod.rs`, `overview.rs`) can read it once that branch lands.
- **Richer drain diagnosis from `Status.details`.** kube 0.97's `ErrorResponse` carries only status/message/reason/code — the drain code already documents this and honestly reports `NotNow`. kube 4 replaces it with a full `Status` (with `details.causes` / `retry_after_seconds`), so the PDB-vs-throttle distinction becomes possible there.

## Looked at, not changed

- **`prefetchQuery` "deprecated".** It is not deprecated in `@tanstack/react-query` 5.102.8 (no `@deprecated` in the installed types). No change.
- **`subscribed` in place of `useLiveQuery`'s visibility refetch.** It would only replace off-screen pause; the machinery that justifies the hook — freshness reporting (`slowed`/`paused`/`everyMs`), the steadiness backoff, the interaction-wake, and refetch-on-return guarded against a herd — all need the interval visible to the hook. Layering `subscribed` on would split "is this polling" across two mechanisms, which is the defect this repo fights.
- **Preserving hand-written comments in `config.toml`.** It is an app-managed, token-bearing file rewritten on every settings change; format-preserving edit adds real risk for marginal benefit, and the recovery above already preserves a file that cannot be parsed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          # The Comment budget step diffs against the merge-base, which a
+          # shallow clone does not contain — `origin/main...HEAD` then exits
+          # 128 and the script that reads it crashes. Full history is what
+          # lets the three-dot diff find where this branch left main.
+          fetch-depth: 0
 
       - name: Setup Tauri build environment
         uses: ./.github/actions/setup-tauri
@@ -71,7 +77,7 @@ jobs:
         # for the rare change that is only about comments.
         if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'comments')
         run: |
-          git fetch --no-tags --depth=1 origin "${{ github.base_ref }}"
+          git fetch --no-tags origin "${{ github.base_ref }}"
           python3 scripts/comment-budget.py "origin/${{ github.base_ref }}" HEAD
 
       - name: Frontend formatting
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1954,7 +1954,7 @@ dependencies = [
  "http",
  "httpdate",
  "mime",
- "sha1 0.10.6",
+ "sha1",
 ]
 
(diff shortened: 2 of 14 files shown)


Dudude-bit:
Also fixes an unrelated bug this PR surfaced: the `Comment budget` CI step diffed `origin/main...HEAD` on a `--depth=1` checkout, where the merge-base is absent, so git exited 128 and the script crashed with a traceback — the first PR to actually run the step. The job now checks out full history, and the script reports the problem plainly instead of stack-tracing.

PR #134, opened 2026-09-07
feat(pod): a Files tab that reads a container and never writes it

belliel:
Stacked on #133 → #132 → #130; retarget to `main` as they merge.

## What is wrong today

Seeing what a container actually has on disk means a shell, and a shell means an image with one. Half the images that matter (distroless, scratch) have none, and even where there is one, `ls` in a terminal is not a list you can sort, filter, tag with the mount a file came from, or download from.

## What changes

A **Files** tab on the pod page. Read-only by construction: the only commands it ever runs are `find`, `stat`, `head` and `cat`, without a TTY, with stdout and stderr kept apart and the exit status read.

- **A ladder, and the rung is named.** GNU `find -printf` first; where that is missing or rejected, a `sh` loop over busybox `stat`; where neither exists, the tab says so and offers the two ways out: a debug container (an ephemeral container sharing the process namespace, reading at `/proc/1/root`, with the listing saying it reads through it) or the pod's declared mounts. `exit 127` is never an empty directory: a test pins it, and the live harness confirms it on a distroless image.
- **Rows stream.** A big directory arrives in batches of 500 over an event stream into a virtualised list, with a running count and a Stop; what arrived stays and says it is not the whole directory. Sorting and the filter apply to what has landed.
- **Every row carries its mount**, from the pod's own volumes: `from demo-config`, with the kind and mount point on hover. The same fact the Connections tab shows.
- **Preview** is text up to 1 MiB with a line count; a binary is refused with the share of non-text bytes that decided it. **Download** goes through the save dialog, byte for byte via `cat`, refused over 100 MiB before a byte moves; nothing is ever written into the container.
- **Bound to the container's life.** A restart after a listing leaves the rows up with a banner saying they are from the previous container, and a button to read the new one.
- A container that is not running says so instead of failing an exec; a refused exec says the cluster refused.

## How to check it

Live, against the k3d cluster (`cargo test --test live_files -- --ignored`): traefik (busybox) lists `/etc` through the `stat` rung with 40 entries and previews `/etc/hostname`; coredns (distroless) answers "no tools", having tried `find` and `sh`, with zero rows; busybox `find` rejecting `-printf` comes back in half a second with exit 1 and its usage text. All three passed here.

Unit, Rust: the runtime's "executable file not found" and a shell's 127 both read as missing, never as success; ordinary failure and success read as themselves; a missing `head` is reported as missing rather than as a failed file; both line formats parse a symlink's target; a stray line on stdout is dropped; binary is a share of the sample and not one odd byte. Frontend: the listing starts at the first mount and tags rows; the no-tools state offers the debug container and the mounts; an empty directory is called empty only after the tool said so; a stopped container is not exec'd into; the debug-container reading is named; a binary preview is refused with its percentage; the mount tagger takes the deepest mount and never a sibling with a shared prefix.

One bug found by the harness and fixed before opening: the listing loop read stderr after stdout, and kube's multiplexer stalls on the small stderr buffer while busybox `find` prints its usage. Both streams are now read together by one task.

Not rendered: no display for the Tauri window here.

## Risk and rollback

Additive: a new tab, five new read-only commands, three new events. Rollback removes them; nothing persists.


Dudude-bit:
Merged main in and fixed the dangerous ones on the branch (80693a8) — holding the merge for the rest.

**Fixed.** The busybox rung exits 0 with no output for a directory it cannot open, so on any busybox image an unreadable or absent path was drawn as a confidently *empty directory*. Reproduced in alpine: `/root` as non-root gave rc=0 and no lines; now rc=2 → `Listing::Failed`. Its `.`/`..` skip also never fired (the short-circuit succeeds on entries that always exist). Pressing **Stop** wiped every row that had arrived and relabelled the partial read as complete — the backend emits `files-done` after a cancel, and by then `stop()` had left `reading`, so the handler took its EMPTY branch. And a failed download **truncated then deleted the host file the reader picked** in the save dialog (same on a merely *unknown* exit) — it now writes to a scratch file beside the destination and renames only on exit 0.

**Still open, worth your call — mostly design-level:**
- With `via` set, the listing always reads `/proc/1/root` of the debug target while the strip labels the rows with whichever container is selected, so container A's files show under container B's name. The debug dialog never learns which container the tab had selected.
- Nothing caps the entry count anywhere — `list_dir` streams every line the tool prints, and the frontend accumulates all of it.
- A cancelled listing still returns `Listing::Listed`, and `FilesDone` has no bit that says partial.
- `read_container_file` has no timeout and no cancel, and preview runs `head` on FIFOs and device nodes.
- Lines the parser cannot read are dropped with no counter, so the "N entries" figure counts only the parsed ones.
- Four `files` count strings interpolate `{n}` into a fixed noun instead of the catalogue plural (Russian needs three forms).

Full list is long (41 confirmed) — happy to paste the rest or take any of them if you'd rather I did.

PR #146, opened 2026-09-08
feat(changes): what changed on a workload, on one clock, with the unwatched hours drawn as gaps

belliel:
Second of the Incident package. Stacked on #145; retargets as the stack merges.

## What is wrong today

At 2 a.m. the question is "what changed", and the app has no page for it. A Deployment's Revisions tab lists ReplicaSets by name and count. What differed between them, who applied them, what Helm did, and what moved while nobody was looking: all of that is a kubectl session away, and "nothing in the last four hours" is an answer the app cannot give because it does not know whether it was looking.

## What changes

A **Changes** tab on Deployment, StatefulSet and DaemonSet, and a **Changes** page under the cluster, both on one clock, newest first.

- **Revisions with the diff.** ReplicaSets for a Deployment; for the other two kinds a new `get_controller_revisions` reads the ControllerRevisions the controller keeps and parses the template out of the patch they carry. Each revision is shown against the one before it: image, env (values, or `secretKeyRef:creds/token` for a reference), requests and limits, containers added or removed, and the checksum annotations a chart bumps to force a roll. The oldest known says it has nothing to compare with. `kubernetes.io/change-cause` is shown under the annotation's own name.
- **Delivery history** through a new `delivery.history` capability, answered by Argo CD (Application `status.history`) and Flux (HelmRelease `status.history`; a Kustomization keeps only its last applied revision and gets one entry). Plain Helm's history comes from the `meta.helm.sh/release-name` annotation.
- **A journal of what this app watched.** Three cluster-wide watches (Deployment, StatefulSet, DaemonSet) run while connected. They record generation, images, desired replicas and config checksums moving, plus creations and deletions, into a persisted per-cluster journal. The rows now carry `generation`, `observedGeneration`, `templateAnnotations` and, for StatefulSet and DaemonSet, `images`.
- **Gaps drawn as gaps.** The watching span is recorded with a heartbeat every 30 s. A crash leaves the span ending at the last heartbeat; a restart closes it there; the timeline draws "Not observed 02:10 to 07:40" as its own row. A relist after a break reports what differs as seen at the relist, behind the gap.
- **`?since=`** marks everything after a moment, which is what the deep-link banner will pass once #127 lands; the tab already reads it.
- **No cause, anywhere.** A test walks every string in the `changes` catalogue, in both languages, and fails on "cause", "because", "led to", "due to", "triggered" and their Russian forms. Two entries an hour apart are two entries an hour apart.

## How to check it

`src/lib/changes.test.ts`: the diff field by field, the gap arithmetic (before the first span, between spans, after the last, an unclosed span ending at its heartbeat), the timeline placing a gap between the entries on either side, and the words test. `src/stores/changeJournalStore.test.ts`: a span left open by a crash closes at its last heartbeat on rehydration and the gap starts there. `src/components/changes/ChangesTimeline.test.tsx`: a gap renders as a note in words, `since` marks rows. Rust: `commands::revisions::tests` on the template parse. Deleting the gap arithmetic fails five tests (checked).

Not rendered in a window: no display on this machine.

## Risk and rollback

The three background watches are new load on every connected cluster: workloads only, cluster-wide, one list each at connect. Everything else is additive. Revert is one commit.


Dudude-bit:
Ran the adversarial review workflow over this branch (6 dimensions, each finding put to independent refutation): 52 survived, ~26 distinct. All of them are fixed in `285fb72`, verified against a kind cluster with the app open.

**Third state, where the page claimed a fact it had not read**

- `unchangedTemplate` said "same template as the revision before" off a diff of four fields. `diffRevisions` now also reads `ports` and `envFrom`, and the sentence states its own reach: *nothing changed in what is compared: image, env, envFrom, ports, resources, checksum annotations*. Which matters most because `revisionOfReplicaSet` set `templateAnnotations: {}` — the checksum diff, the whole point of a config-only roll, never fired for Deployments at all. `ReplicaSetInfo` carries the pod template's annotations now.
- A `ControllerRevision` whose snapshot will not parse arrived with empty containers and was drawn as a revision that ran nothing, every container "removed". `template_read` on the wire, `templateKnown` on `Revision`, and the pair is reported as *not known* rather than compared.
- `envWord` spelled every downward-API variable `fieldRef:null` — one word for `spec.nodeName` and `status.podIP`, so a change between them was no change, and an added one printed the string "null". Composed from the field that identifies the source now.

**The gap machinery, which was under-claiming in four ways**

- A relist left the span open, so a break read as watched — and because `failed` dropped the kind from `synced`, the relist diff was suppressed in exactly the case where the gap did get drawn. `restarted` closes the span, the baseline is kept across the break, and the entries the relist finds carry `atRelist` so `journalSeenAtRelist` (written for this and never rendered) appears beside them.
- `event-bridge-lagged` was not listened for at all. It now blinds every watch and re-subscribes.
- A missed heartbeat — a sleeping laptop, a frozen tab — was taken as proof the watch was alive the whole time. A tick more than three intervals late closes the span at its last `seenAt` and opens a new one.
- And the opposite: `gapsOf` ended an open span at `seenAt` while the window ended at `now`, so a "Not observed" box blinked at the top of the list on a 30-second cycle against a cluster being watched perfectly.

**One fact, two readers**

- The change cause for StatefulSet/DaemonSet was read from the pod template. Both controllers copy the workload's annotations onto the revision itself; `kubectl` denormalises it onto the template only to print it. The Deployment half was already reading it correctly from the object.
- Journal images were paired by position while `diffRevisions` keyed by name, so removing a container invented changes on every container after it. `StatefulSetInfo`/`DaemonSetInfo` carry `containerImages` (name + image) and the sentence names the container.
- Helm history was cached under a private key that the Helm page's rollback invalidation cannot reach; it uses `["helm-history", name, namespace]` now.

**Also**

- A `claimed` delivery — the object carries the label, the owner does not list it — had its commits drawn as this object's own history. Claimed owners are named in the unread list instead.
- `deliveries.error` was the one of four reads the tab stayed silent about: a 403 on `list applications` read as "nothing delivers this".
- Three cluster-wide watches made the feature dead for a namespace-scoped reader; asked per namespace now.
- A rollback re-adopts the existing object, so the newest revision carries the oldest creation time. Marked, rather than drawn as having happened before the change it undid. Revisions the cluster no longer holds between two it does are counted.
- The 24h/7d buttons printed their keys in every language. Flux's Kustomization history returned the app's own word "applied" into a field that carries the cluster's.
- localStorage was rewritten whole on every watch batch and every heartbeat; writes are batched and spans are pruned and merged. The timeline caps at 300 rows and says how many it did not draw. The tab's three reads go through `useLiveQuery` so they refresh while it is open.
- The context name is not a cluster: a rebuilt `kind` under the same name inherited the old one's watched hours. Checked against the `kube-system` UID, and a reader who cannot read it loses nothing — "could not tell" is not "another cluster".
- `changes.test.ts`'s "never call a change the cause of anything" guard could not match a single Russian word: `\b` in JavaScript is an ASCII boundary, so the half of it reading `ru.changes` had never run.

Gates: tsc, lint, 2613 frontend tests, 608 Rust tests, clippy, comment budget 4%. Every new branch has a test that fails when the branch is deleted.

PR #147, opened 2026-09-08
feat(logs): one stream over every pod of a workload, a lane per pod, coverage stated

belliel:
Third of the Incident package. Stacked on #146; retargets as the stack merges.

## What is wrong today

A Deployment's Logs tab reads one pod at a time from a dropdown. A rollout, a scale, a DaemonSet on twelve nodes, a CronJob's last three runs: the question is always "what did all of them say, in order", and the answer was twelve tabs and a stopwatch. StatefulSet, DaemonSet, ReplicaSet, Job and CronJob had no Logs tab at all.

## What changes

The log viewer learns a second pane: many pods, one buffer, a lane per pod.

- **One stream per container of every pod**, interleaved by each line's own timestamp inside the existing reorder window, not by arrival. The stream hook now takes `sources` (pod, namespace, container) and a `paneKey`; a change in sources under the same pane stops the streams of pods that left and opens the new ones with their own tail, and never clears the buffer. Deleting that delta fails two tests: a rollout would drop the old pod's last words.
- **Lanes.** The gutter colour and the legend chip are the pod. A line can name its pod by colour alone, by the last characters of its name (what a colleague reads off a line pasted into a chat), or in full. A pod the list no longer has stays as a grey lane, struck through, with its lines kept.
- **Coverage, in words**: "3 of 3 pods streaming · 1 refused · 1 gone, lines kept", next to what a lane stands for under this kind: pod for Deployment, ReplicaSet and Job; ordinal for StatefulSet; node for DaemonSet; run for CronJob.
- **A refused stream is a number in the coverage and a mark on its chip**, and a banner with the pod's name. A pod that ended is not a banner on a workload pane: rollouts end pods all the time.
- **Download** is one file per pod and container, the last ten thousand lines of each, from the API; a pod that is gone has no log left to ask for.
- `pod=` is a filter term and a field the legend counts by; a single-pod pane does not offer it.
- The Logs tab now exists on Deployment (the dropdown is gone), StatefulSet, DaemonSet, ReplicaSet, Job and CronJob. The pod pane and the peek panel are unchanged: same component, `podName` and `containers` as before.

## How to check it

`useLogStream.test.ts`: a rollout keeps the old pod's line, stops its stream, opens the new pod with a tail and registers listeners once; a refused pod is named with its container. `LogViewer.test.tsx`: one chip per pod, the coverage sentence, a replaced pod kept as a gone lane with its count, the short-prefix label, a refused stream counted, an empty workload said in words. The 198 existing log tests pass unchanged.

Not rendered in a window: no display here. Ordering by timestamp and the reorder window are the existing code, exercised by the existing tests.

## Risk and rollback

The single-pod pane goes through the same hook with `containers` mapped to sources; every previous test on it passes. A workload with many pods opens many streams at once, bounded by the shared cap. Revert is one commit.


Dudude-bit:
Ran the review workflow over this branch (6 dimensions, each finding put to independent refutation): 43 survived, ~25 distinct. All fixed in `0c681da`, verified against a kind cluster with the app open.

**`pod=` was a term on one side of the boundary only.** `matchesTerm` gained a `pod` arm, the field index gained `visit("pod", …)`, the lane label and the row detail both mint a `pod=` chip — and `QueryTerm::Field` in `filter.rs` still special-cased only `container`. Flip that chip to intake and Rust looks `pod` up in the parsed fields, finds nothing, and every stream of every pod goes silent while the buffer sits there reading as "the workload stopped writing". `shared/log-query-conformance.json` had no `pod` case, so nothing failed: deleting the whole TypeScript arm left 989 tests green. Both evaluators now have it and four corpus cases hold them together — each side sabotaged and caught.

**A pod list that could not be read was drawn as a workload with no pods.** `pods = []` from a 403 and `pods = []` from an empty workload arrived at the pane as the same value, and it said "No pods to read from yet." — on all six pages, while the Pods tab beside it drew the refusal from the same query. Job and CronJob were swallowing the error in the queryFn on top of that. `podsError` travels now, and the test that asserted the empty sentence for `pane([])` has a sibling that fails if a refused list says it.

**Collapsed runs straddled pods.** `groupKeyFor` folds container and level so a run "can never straddle either" — and a workload pane put five replicas' identical lines through it. Five pods saying `connection refused: db:5432` inside one reorder window collapsed into one row, drawn in the head pod's lane, while the legend beside it said each of the five wrote one line. `pod` is in the key now, and it is visible on screen: two replicas writing the same sentence in the same second are two rows in two colours.

**The stream lifecycle**

- Reconnect on a workload pane cleared the buffer — every other pod's lines, including pods that had left, whose streams are gone and cannot be read again. Per-lane now; the session restart stays for a pane-wide change.
- A source whose open was rejected was dropped from `opened`, so the next `sync` — every rollout tick — silently retried it and stacked another identical failure. Two pods, one in ImagePullBackOff, read "1 of 2 pods streaming · 4 refused". Kept in a `refused` set now, deduped, and cleared when a retry succeeds.
- The opposite for a container that had not started: the refusal was permanent for the life of the pane, beside a pod list showing it Running. The source carries a coarse `started` bit, and turning true releases it.
- A pod joining mid-session opened with `tailLines: 1250`. The buffer is append-only and everything in it is newer than that pod's history, so the clock jumped backwards mid-pane and a following reader was dragged into minutes of old lines — six times on a six-replica rollout. It streams from here now. Watched live: scaling 2→3 put the new pod in at `request_id=req-5` with the timestamp column still monotonic.
- `isConnecting` was never cleared for a pane with no sources, so a workload with no pods yet sat on the skeleton with a dead toolbar.

**The coverage sentence counted the wrong things.** `gone` is what the streamer emits when a followed stream reaches EOF — the ordinary end of every init container and of every pod of a finished Job — so a Deployment with one migration init container read "0 of N pods streaming" while all of them were writing. `refused` counted container streams inside a sentence whose other clauses count pods. And `broken` is the could-not-look state, not a refusal: it says "{n} pods could not be read" now, and a paused pane says it is paused rather than claiming a gap.

**The lanes**

- Colours came from the position in the live pod list, so one pod dropping out shifted every pod after it and the buffer repainted itself mid-rollout — with the gutter being the only identity in the default label mode. A first-seen order now, and the live check shows two lanes keeping their hues across a join and a departure.
- `pod` and `container` inherited the field index's 50-value cap, so past fifty pods every lane chip's count read zero and departed lanes vanished with their lines still in the buffer. A 60-node DaemonSet crosses that on its first read.
- A Job with `parallelism: 3` owns all three pods, so `laneRule="run"` drew three identical chips; same for a retried CronJob pod. Job uses `pod`, and a rule-name shared by more than one lane falls back to name/suffix.
- Pods that exist only in a history read got no chip, no colour and no count — which is exactly what the workload scope is for.

**Also:** the failure banner was worded from the first container of that name across every pod — another pod's exit code and restart count; a pane whose only failures were `gone` rendered a completely blank body; the download stopped at the first unreadable pod; "Every container is hidden" on a pane that hides pods, with `everyLaneHidden` written and rendered nowhere; `laneLabelHint` promised the pod name under all four rules; the seven `Filter on x=y` tooltips were raw English (five of them predating this branch); and `sourcesKey` joined on a literal NUL byte, which makes grep call the file binary — a real cost in a repo whose rule is to grep every reader of a fact.

Gates: tsc, lint, 2634 frontend tests, 617 Rust tests, clippy, comment budget 13%. Every new branch has a test that fails when the branch is deleted.

PR #149, opened 2026-09-09
feat(pod): most likely, said from the chain the app read, with a search and a hand-off that carry no secret

belliel:
Fourth of the Incident package. Stacked on #147; retargets as the stack merges.

## What is wrong today

A pod in CrashLoopBackOff shows the reason the kubelet wrote and stops there. The next four steps are always the same and always by hand: read the last lines of the previous run, notice the address they name, find out whether that address is a Service, and whether anything is behind it. At 2 a.m. that is four screens and a guess.

## What changes

A **Most likely** panel on the pod's Overview, above the facts, for a pod with a problem the app can read a chain for.

- **The trouble, in priority order**: a container that cannot stay up, one killed for memory, an image that will not pull, a volume that will not mount, a pod nothing will schedule, a probe the kubelet keeps failing.
- **The chain, read and never probed.** For a crash loop it reads the last 40 lines of the previous run, takes the last address a failed connection named, and classifies it: this pod itself (a sidecar), inside the cluster (a Service, and how many endpoints are ready), or outside (where the app can see the road and not the far end). The sentence changes with what came back: "That address is Service shop-db-rw, which has nothing ready behind it right now; the pod itself is probably fine."
- **Every guess is hedged.** A catalogue test walks every `guess*` string in both languages and fails one without "probably"/"usually" (or their Russian forms). Under the sentence, in words: nothing here sent a packet.
- **Checks, not advice**: links into the app (the Service, the ConfigMap and Secret the pod mounts, the node, the container template, the log tab), each one a place to look.
- **What was not read is named**: a refused log read, a refused Service list, endpoints that did not answer, and NetworkPolicies, which this app has no reader for at all.
- **Search it** opens the engine with the reason and the exact message, with pod, namespace, image, node and host names replaced by `…` (a switch; off for people whose names are not sensitive), and `utm_source=rubick.tech`.
- **Copy for agent** is plain text: the pod, its containers and exits, the log lines, folded events, what the app read of the chain, the mounted ConfigMaps and Secrets **as names and key counts**, a "Not read" list, and the app's own sentence labelled as a guess. A Secret value cannot get in: nothing in the report's input accepts one, and a test hands it an object carrying values anyway and reads the output for them.
- **Settings › Search and hand-off**: engine (Google, DuckDuckGo, or any URL with `{q}`), strip names, include log lines, show the panel at all.

## How to check it

`src/lib/hints.test.ts`: the trouble priority, the address classification (sidecar, in-cluster, outside, and a clock that is not a port), the sentence per chain shape, the query with and without stripping, and the report that never carries `hunter2`. `MostLikelyPanel.test.tsx` renders the panel against a crash loop whose log names a Service with nothing ready, and draws nothing for a healthy pod or when the setting is off.

Not rendered in a window: no display here.

## Risk and rollback

The panel only reads: logs, Services, endpoints. All three are separate queries whose failures become lines rather than a missing panel. Off by one switch. Revert is one commit.


Dudude-bit:
Ran the review workflow over this branch (6 dimensions, each finding put to independent refutation): 51 survived, 11 of them high. All fixed in `c1dd367`, verified against a kind cluster with a crash-looping pod.

**Two of them are about data leaving the machine, and both are now closed.**

*Search it* sent the **raw log line** to the search engine. Stripping replaced five of this pod's own names and nothing beside them, so a line like `PSQLException: Connection refused ... url=jdbc:postgresql://db.shop.svc.cluster.local:5432/app?user=svc&password=Hunter2-prod` reached google.com with the password in it — with the switch the settings describe as the cautious one turned **on**. The stripping was also order-dependent: `shop` was replaced before `db.shop.svc.cluster.local`, after which neither the host rule nor the `.svc` rule could match what was left. The query now carries what the app *recognised* — `connection refused`, `i/o timeout` — never the line; the shapes a secret takes are removed whatever the switch says; and the name pass runs after the regexes, longest name first.

*Copy for agent* copies 40 raw lines of the previous run while three places promised otherwise — the doc comment, the log-lines setting ("Secret values are never included, whatever this says") and the toast ("secrets never"). The guarantee is earned only by the mounts block; a Spring Boot pod that prints its resolved `spring.datasource.url` at startup put the production password into a chat under that sentence. The test that appeared to hold it passed a `values` field that `MountedConfig` does not declare and the function never reads — it could not fail. The lines now go through the same redactor, the copy says what it actually is, and the test hands it a DSN and a JWT and reads the output for them. Verified through the real app: the clipboard shows `...app?user=svc&password=…` with the rest of the line intact.

**The third state, in a panel built on reads that can be refused**

`Chain.service` had no known bit, so a 403 on the Services of the namespace produced *"no Service in this namespace answers to it"* — a claim about the cluster from a read that never happened, and one that sends a reader hunting for a wrong address. `servicesKnown` and `endpointsKnown` default to false, a Service found with its endpoints unread is reported as uncounted rather than as empty, and there are catalogue strings for both. A refused *events* read used to hide the panel outright, which is byte-identical to a pod with nothing wrong — three of the six troubles are read from events; it now draws with the not-read line and no guess.

**What the chain actually proves**

- `127.0.0.1` is classified `sidecar`, and with no container claiming the port the arm fell through to the outside branches: a pod dialling its own loopback was told *something outside the cluster refused the connection*. It has its own sentence now, and the sidecar is the container whose `ports` include that port rather than the next one in the list.
- The Service and sidecar headlines hardcoded "refused the connection" whatever the line said, and `no route to host` — neither a refusal nor a timeout — was reported as a refusal, then followed by "a firewall would have timed out". Three-way now.
- `db.shop`, the cross-namespace form, was called outside the cluster: it gated off the Service lookup and added a NetworkPolicy line about a hop that never leaves. And `shop-db-rw.billing.svc.cluster.local` was matched on the bare name against *this* namespace's Services, so a same-named Service next door was reported with its endpoint count. The namespace segment is read now, and a namespace the app did not list is named in "not read".
- `main.py:42` matches the host:port shape exactly and the last match on the line won, so a stack-trace frame became the address the whole chain was read from.
- A container OOM-killed under `restartPolicy: Always` spends its life in CrashLoopBackOff, and the OOM arm sat *after* the crash-loop arm — unreachable for exactly the pods it was written for, while the Containers tab on the same page said OOMKilled. An OOM from days ago, and an hour-old FailedMount on a pod Running since, both put a permanent trouble panel in the present tense on a healthy pod.
- The OOM headline printed `1073741824` and called it this container's limit; it is the pod's total, `ContainerInfo` carries no resources, and it says so now, in words.
- The probe hint invented a container called `app` and a probe called `readiness`. A single-container pod answers the first; otherwise both have strings that name neither.
- Every mounted Secret was offered as "a pull secret the pod mounts" — `imagePullSecrets` is a different field this app does not read.

**Words and counts.** Eight counted sentences were flat strings, so English said "1 restarts" and Russian got one form for every number. Each counted noun is a plural object now, chosen as its own sentence and interpolated — the `Saying`-in-`Saying` pattern the Events stories already use — because no language can hand another a substring of its own plural. Verified on screen: *«app завершился с кодом 1, уже 6 перезапусков»*. `stateWord` composed English fragments ("exited 137", "running, not ready") that landed inside Russian sentences. A custom search engine accepted any string including an empty one, after which the line under the button read "opens ; change the engine in Settings" and the query went to the clipboard instead of a browser.

**Tests.** The name-stripping test proved nothing about names — deleting the whole stripping loop left the suite green; no test covered any "not read" line, the third state the panel exists to show; and `troubleOf`'s failedMount and probeFailed arms could both be switched off with everything passing. Every branch added or changed here has a test that fails when the branch is deleted; I verified that one at a time.

Gates: tsc, lint, 2674 frontend tests, 617 Rust tests, clippy, comment budget 8%.

PR #150, opened 2026-09-09
feat(tell): did it work, answered only once the object confirmed the action, or timed out in words

belliel:
Last of the Incident package. Stacked on #149; retargets as the stack merges.

## What is wrong today

Restart, Scale, Apply and a changed image all end the same way: a toast saying the request was accepted, and then silence. Whether the rollout finished, gave up, or is still going is the reader's job to go back and check, and the object looks exactly the same for the first second after the click as it did before it.

## What changes

Every one of those actions is followed for two minutes and answered, through the watch machinery #130 already built.

- **Nothing is said until the object confirms the action reached it.** A restart, an apply or an image change is acknowledged by `metadata.generation` moving past the one the page held before the click; a scale by `spec.replicas` becoming the number that was asked for. A Deployment that was settled before the click stays silent, which is the whole point: "rolled out" on the first tick would be a verdict about the state before the action.
- **Three answers, never a fourth.** Rolled out, carrying what it saw ("3 of 3 ready, revision 9"); failed, carrying the controller's own words off `Progressing=False`; or, at two minutes, **no answer**, carrying the last look. A silence is an answer with a name rather than a watch that sits open for a day.
- The Watching tab says which action a watch is following ("rollout · after scale to 5") and paints a timeout amber, apart from a failure.
- Asking again about the same object replaces the open watch: the new click is the one whose outcome the reader wants.

Wired at every call site that mutates a workload: the Deployment page (scale, restart, image), the StatefulSet page (scale), the shared object actions used by lists and the peek panel (restart of a Deployment, scale), and the YAML editor's apply for a workload kind. A restarted **pod** is deliberately not followed: that is a deletion, and the pod that replaces it is a different object.

`StatefulSetDetailInfo` and `DaemonSetDetailInfo` gained `generation` and `observedGeneration`, which is what "the controller has caught up" is read from.

## How to check it

`src/lib/tell-me-when.test.ts`: a settled look at the old generation says nothing; the verdict lands only once the new generation is observed and settled; a scale is acknowledged by the count when the page never knew a generation; a failed rollout carries `ProgressDeadlineExceeded`'s message; the last look is remembered in words. `src/stores/tellMeWhenStore.test.ts`: a watch past its deadline closes as timed out with that last look, once. Removing the acknowledgement rule fails four of them.

Not rendered in a window: no display here.

## Risk and rollback

Additive on top of #130's watches, and bounded by the same cap of 12 per cluster; a followed action expires in two minutes rather than a day. Revert is one commit.


Dudude-bit:
Reviewed with current `main` merged in (17 commits behind; clean). One fix pushed; the other finding is a number you should choose rather than one I should.

**"Rollout failed" was read off the previous rollout's condition.** The apiserver bumps `metadata.generation` the moment the action lands, while `status` still describes the rollout before it. So: a Deployment stuck on `Progressing=False / ProgressDeadlineExceeded`, the reader clicks *Change image* to fix it, `acknowledged()` passes on the generation bump, and the very first `applied` event — the watcher's own initial snapshot — answers `rolloutFailed` with the **previous revision's message**. Red dot, toast, OS notification, within a second of the click. `settle()` then closes the watch, so the fix that actually succeeds half a minute later is never reported.

What makes it a clear defect rather than a race is the asymmetry: the success arm already refuses a stale *yes* — it requires `observedGeneration >= generation` — and the failure arm two lines above applies no test at all. The code refuses to believe a stale "it worked" and believes a stale "it broke".

`caughtUp` alone would not have fixed it, which is why I did not just add it: once the controller observes the new generation, `DeploymentTimedOut` re-asserts `Progressing=False` before anything has had time to progress. The bit that settles it is the one the cluster writes — `lastTransitionTime`. `After` now carries the moment of the click (stamped in `useAsk`, one place), `Rollout` carries the condition's transition time, and a failure is believed only when it transitioned at or after the ask. Where a cluster leaves no stamp it falls back to the success arm's guard, which is still better than believing whatever was there.

The existing test uses `look(5, 5, …)` — generation equal to observedGeneration — so neither window was covered; there is a case for both directions now.

**Not fixed, and yours to decide:** `OUTCOME_DEADLINE_MS` is two minutes, while Kubernetes' own `progressDeadlineSeconds` defaults to 600s. At two minutes the watch is marked done, which drops it from `openIds`, which unsubscribes the stream — so the app stops looking. Any rollout slower than two minutes (a few-hundred-MB image pull, `minReadySeconds`, a StatefulSet going one pod at a time) is told "no answer within two minutes" and then never told the real answer, which lands on a stream nobody is holding. Before this PR the same watch had the day-long TTL and would have said "rolled out" whenever it happened — so the promise is weakest for exactly the rollouts slow enough to be worth walking away from. Reading the object's own `progressDeadlineSeconds` would be the honest bound; the catalogue sentence hardcodes "two minutes" as well.

Dudude-bit:
Second pass on this branch, three more fixes pushed.

**`caughtUp` had no test that fails when it is deleted.** The existing cases all used `look(5, 5, …)` or replica counts that were unsettled anyway, so both arms passed with the guard removed. There are two now: a look that is fully settled while `observedGeneration` still trails, and a failure with no `lastTransitionTime` where the fallback is the only thing holding it. Both fail on sabotage; the failure one distinguishes *which* look answered by giving the two conditions different messages.

**`LOST_SIGHT_MS` and `OUTCOME_DEADLINE_MS` are both two minutes, and they raced.** A watch whose stream went down reaches its deadline at about the moment it would have said so, and whichever timer wins decides what the reader is told. When `timeOut` won, the answer was "no answer within two minutes" — said about a window nobody watched — instead of "lost sight of it". The app knew it could not look and said something weaker. `outOfTimeVerdict()` now picks the verdict from the watch's own state, and both the store and the hook call it so they cannot disagree; a watch already told it lost sight is closed without a second notification.

**`detail` carried both the cluster's words and ours.** Its doc comment says "the cluster's own words", and for a rollout it was `seenWords()` — "3 of 3 ready, revision 8", composed in English at judge time and printed unchanged to a Russian reader by both the Watching tab and the OS notification. It is a `Saying` now, resolved at render by `detailWords()`; a controller message is still a string and still goes out whole, and the type tells the two apart. Checked live in Russian: the toast reads «cloud-demo раскатился · готово 2 из 2, ревизия 2».

The `progressDeadlineSeconds` question from my last comment still stands and is still yours.

PR #158, opened 2026-09-10
feat(services): a home page of the workloads somebody pinned, and nothing else

belliel:
Stacked on #153 (`feat/share`). Review the last commit; the rest is the stack underneath.

## What is wrong today

Ask somebody which services they are responsible for and they can name four. The app cannot: it offers Deployments, StatefulSets, DaemonSets and CronJobs as lists of everything in a namespace, ordered by name. Every morning the same person filters the same list to the same four things, opens each one, and then opens its Service, its Ingress, its events and its rollout to find out whether anything happened overnight.

Nothing on the home page is about *their* services, and no existing screen can become that, because every list is derived from the cluster rather than from the person.

## What changes

A first block on the home page holding the workloads somebody pinned by hand, and nothing else.

**Membership is explicit and stays that way.** A card is there because a person pressed Pin on the object's page, and it goes when they press it again. Nothing infers membership from what was opened recently, from a label convention, or from what looks important. A test asserts the page shows nothing unpinned, so the rule cannot erode into a heuristic later.

**One card answers the five questions that otherwise cost four screens:**

- what state it is in, `3/3 ready`;
- the way in, as an address to paste where the chain knew a host, otherwise the Service and its ports;
- what changed last, from the journal this app already keeps;
- what it is still waiting on, from the open "tell me when" questions;
- what nobody looked at, from the read's own `notLookedAt`.

**Two answers a replica count would have swallowed.** A service somebody deleted reads `does not exist`; one this token may not read reads `could not read`. Drawn as `0 ready`, the first tells a person their service is down when it is not there, and the second tells them it is down when nobody looked. `stateOf` keeps the three apart and a test fails if any two are merged.

**Pins are per cluster**, not per namespace scope: what is mine stays mine when I narrow the scope to look at something else. Twelve per cluster, because each card is one neighbourhood read; past the cap the app refuses and says so rather than dropping the oldest thing somebody chose to watch.

### Two facts got one reader instead of two

- `Journal`, the switch that puts a journal entry into words, is now exported from `ChangesTimeline` and used by the cards. A second switch over `field` is exactly how the changes page and this one would come to disagree the next time a field is added.
- `ASK_SHORT` moved from `WatchingTab` into `tell-me-when.ts`, beside the `Ask` union it is keyed by. A new ask now fails to compile in one place instead of being silently unnamed on one of the two surfaces.

### What it costs

Nothing new on the backend, and no new query key. `useConnections` is the same query the object's own detail page makes, with the same key, so a pinned service's card and its page share one answer and opening the service from the card costs no request at all.

## How to check it

- `bun run test` — `src/lib/my-services.test.ts` (12), `src/stores/pinnedServicesStore.test.ts` (4), `src/components/services/MyServices.test.tsx` (4).
- The rule itself: `MyServices.test.tsx` mounts the panel with a pin belonging to another cluster and asserts no card is drawn.
- By hand: open a Deployment, press Pin, go to the home page. Switch context and the block is empty with the sentence that says how to fill it. Press Pin on thirteen things in one cluster and the thirteenth is refused by name.

Rendered from source only, not opened in the app: the layout was checked as a static page in both themes before it was built, and the two are the same markup. [The mockup](https://luh3a60l1qci.postplan.dev) is the light and dark states side by side.

## Risk and rollback

Additive. Nothing existing reads `pinned-services`, and an app with no pins renders one empty-state box. Reverting removes the block and the four Pin buttons; the two extracted readers (`Journal`, `ASK_SHORT`) would revert with it and are behaviour-identical to what they replaced.

## Summary by CodeRabbit

- **Новые возможности**
  - Добавлено закрепление сервисов в выбранном кластере с ограничением количества закреплений.
  - На странице обзора кластера отображаются закреплённые сервисы, их состояния, точки входа, изменения и непросмотренные вопросы.
  - Добавлены кнопки закрепления для CronJob, DaemonSet, Deployment и StatefulSet, а также возможность открепления.
- **Локализация**
  - Добавлены русские и английские переводы для закреплённых сервисов, их состояний и пустых состояний.
- **Тесты**
  - Добавлены проверки закрепления, отображения сервисов, фильтрации по кластеру и обработки различных состояний.


Dudude-bit:
Rebuilt on current main and reviewed. Its base was `feat/share`, which was the tip of an old stack; this PR's own commit is replayed on main and the base is retargeted, so what you see is the feature and nothing else. (GitHub closed the PR when the branch stopped descending from its old base — reopened.)

**What a card claimed that it did not know.** The read is kept across refreshes, so a workload deleted or refused after the first good answer left that answer on the card, green, with the failure beside it unread. "Client not found" and "Plugin not found" — this app failing, not the cluster answering — were both read as a deleted workload, which tells somebody to rebuild something that is still running. A CronJob has no replicas to be ready, and the backend's 0 of 0 read as "all ready". And "could not read" was drawn for the first second of every visit, before anything had answered.

**The ways in.** The object hop and the published hop are the same Service under the same key, and only the published one knows whether anything is behind it — first-one-wins dropped that answer every time, so "nothing behind it" could never be drawn. Meanwhile "nothing publishes this" was stated over a Services list the cluster had refused, which `notLookedAt` names.

**Smaller:** `waitingFor` matched a watch on context, namespace and name but not on the kind, so a Job's wait showed on the Deployment of the same name — both its neighbours match on all three. And the rebuild had carried another branch's whole `share` catalogue section into this one, in both languages, with no reader here.

Each fix has a test that fails when its rule is deleted.

Gates: tsc, lint, 3090 frontend tests, 721 Rust tests, clippy, both formatters, comment budget 9%.

Dudude-bit:
A second pass over what the review left.

**"What is pinned here" was written three times.** The store carried `forContext` and `isPinned` that nobody called, and both screens filtered the list themselves — the shape where one gets fixed and the other keeps the old answer. One function now, with the cluster scope and the oldest-first order under test.

**Twelve neighbourhood reads every eight seconds.** Each card asks for a whole neighbourhood, so a full page was ninety reads a minute for a screen somebody glances at over coffee. The cards are on the 30-second rate, and a test holds the cap and the rate against each other so moving either is noticed.

**Nine rules could be deleted with the suite green** — the CronJob guard, the not-a-workload guard, the check that a 404 is about *this* object, the pod-name shape, the endpoints `whole` bit, the kind in the journal match, `isOpen`, and both orders. Each has a test that fails when the rule goes; found by deleting them one at a time rather than by reading.

PR #159, opened 2026-09-10
feat(alerts): paste the alert you were woken by, open the object it names

belliel:
Stacked on #158 (`feat/my-services`). Review the last commit.

## What is wrong today

At four in the morning an alert arrives in Telegram. It is read by eye, the context is switched by hand, the pod is found in a list by hand, and the time window is typed in by hand. Three retypings of one name, by somebody who was asleep ninety seconds ago.

Nothing in the app helps, because the alert is text and the app only takes names.

## What changes

Paste the whole message into the palette. It is caught on `paste` rather than read out of the field, because an `input` drops the newlines on the way in and the newlines are the grammar. Nothing opens on arrival; `esc` puts the search back.

### The label key is the kind, and that is not a guess

The kubernetes-mixin, which is what kube-prometheus-stack ships, names the object in the label key: `namespace`, `pod`, `container`, `deployment`, `statefulset`, `daemonset`, `job_name`, `horizontalpodautoscaler`, `poddisruptionbudget`, `persistentvolumeclaim`, `persistentvolume`, `node`. So this is a lookup table, and it reaches every kind that already has a page rather than only pods.

**Two traps it exists to avoid, both of them on ordinary alerts:**

- `job = kube-state-metrics` is a Prometheus scrape job. The Kubernetes Job is `job_name`. Reading the first as the second opens an object that does not exist, on every alert this stack sends. `instance`, `endpoint`, `prometheus` and `service` are the same; they are listed as seen and dropped rather than silently ignored.
- A `KubeDeploymentReplicasMismatch` carries a `pod` label, and that pod is kube-state-metrics. So the alert's own name picks the subject, and every other object it names is offered underneath as `also named`. Opening the wrong one sends a person to read the monitoring stack's logs while their checkout is down.

The name is a preference, not a rule: `KubePersistentVolumeFillingUp` is named for a volume and labelled with a claim, so a preference nothing satisfies falls through to what the alert really carries.

### One grammar, not four parsers

Alertmanager's plain-text body is `Labels:`, ` - name = value` lines, `Annotations:`, `Source:`, and email, Telegram, OpsGenie, MS Teams and Mattermost all render that same body. Grafana's is the same shape plus `Value:`, `Silence:`, `Dashboard:` and `Panel:`, so it is a superset. PagerDuty and Slack are wrappers whose own text carries an incident number and a service name, with the Kubernetes identity in the description they were handed, so the chrome is stripped and what is inside is read. Datadog is a second key table, not a second parser: `kube_namespace`, `pod_name`, `kube_cluster_name`, `kube_deployment`, `kube_container_name`.

### Where each field came from, and what was only a guess

Every recognised value names the line it came from, because a wrong read is only cheap while it is still on this screen.

- Alertmanager's subject template renders the extra labels as **values with the keys thrown away**, and that one line is what most Telegram templates send. Anything recognised there is recognised by shape, says `by shape`, and the values it cannot place are offered as candidates rather than filled into fields.
- Where no label names a cluster, the `Source:` host is offered as a guess and marked as one. A cluster is settled only by an exact context name: two of yours whose names both contain the alert's word is exactly the case where guessing opens a healthy copy of the thing that is down.
- A time with no date is refused. Dating `03:14 UTC` means guessing a day, and a window opened on the wrong day shows a calm hour that reads as "nothing happened".

### The claim stays a claim

Opening from an alert puts a banner on the object's page: the alert's sentence, quoted and dated, beside what this app read itself. A pod that recovered eleven minutes after the rule fired is the ordinary case. Where the page has not read yet the banner says so, and where the read was refused it says that, rather than letting the alert's words stand unanswered as the object's state.

The banner is mounted once, in `ResourceDetailLayout`, so it covers every kind an alert can name instead of being copied into six pages that then drift.

## How to check it

- `bun run test`: `src/lib/alerts.test.ts` (21), `AlertReadingPanel.test.tsx` (5), `AlertBanner.test.tsx` (4).
- The corpus is the shapes real senders render: an Alertmanager body, its subject line alone, a Grafana body, Datadog tags, and an alert wrapped in PagerDuty.
- Sabotage check on the `job` trap: adding `job: "Job"` to the kind table fails three tests, including the Grafana one that must find no Kubernetes object at all.
- By hand: copy any Alertmanager message, `mod+k`, paste.

Rendered from source only, not opened in the app. The layout was checked as a static page in both themes first; [the mockup](https://f8l0qzx52hqu.postplan.dev) has every state, including the two ambiguity cases.

## Risk and rollback

Additive. `looksLikeAlert` is deliberately strict and requires a sender's own mark (`[FIRING]`, `[Alerting]`, `alertname =`, a `Labels:` line, `Triggered:`, `startsAt`, `kube_namespace`), so an ordinary paste into the palette still searches; a test pins five everyday inputs as not-alerts. Nothing is sent anywhere: this is a paste box and a parser, with no integration and no polling. Reverting removes the panel, the banner and the store.

## Summary by CodeRabbit

- **Новые возможности**
  - Добавлен разбор алертов Alertmanager, Grafana и Datadog.
  - Командная палитра показывает сведения об алерте и позволяет выбрать кластер, namespace и объект.
  - Найденный ресурс можно открыть напрямую, включая вкладку изменений.
  - На страницах ресурсов отображаются текст, статус чтения, время и ошибки алерта, в том числе при загрузке.
  - Добавлена поддержка русского и английского языков.

- **Тесты**
  - Расширены проверки разбора алертов, выбора целей, отображения баннеров и навигации.


Dudude-bit:
Rebuilt on current main (its base was `feat/my-services`, itself on the old stack) and reviewed.

**The banner was mounted below the layout's two early returns.** So on an object this cluster does not have — the arrival the whole feature exists for — and on one still being read, it showed nothing at all: `error` reaching it could only ever be `undefined`, and its own "could not read" branch was unreachable code. It is above both now and says which of the three states the page is in.

**The cluster was thrown away between the panel and the banner.** `AlertTarget` carries it, `openFromAlert` uses it to pick a tab, and then only `{kind, name, namespace}` reached the store — so an alert about `payments` in one cluster was drawn on `payments` in another, beside that one's green "read just now".

**A cluster-scoped object could never match.** The alert's `namespace` label was stored for a Node or a Namespace too, and their pages have none, so the banner never appeared on exactly those arrivals — including the panel's own "or the namespace" button.

**`esc` closed the whole palette.** The Open button's `autoFocus` took focus off the palette's input the moment the panel mounted, and the handler that puts `esc` back on the search box lives there; ⌘K then reopened the palette on the stale alert instead of the search.

**And two about words:** ten detail pages have no status badge — a ConfigMap has no state to show — and the banner read that as "This app is still reading it", a claim about the app that never went away; every severity was drawn in the red reserved for a failure, so an `info` alert arrived looking like something that wakes somebody.

The rebuild had also carried two sibling branches' whole catalogue sections into this one, in both languages, with no reader here; removed.

Gates: tsc, lint, 3101 frontend tests, 721 Rust tests, clippy, both formatters, comment budget 11%.

Dudude-bit:
A second pass over what the review left, confirmed with the app open.

**A guess was reported as a label.** A cluster read out of a `Source:` hostname was drawn as "no cluster of yours is called that" — crediting this app's guess to the sender. And a value read off `deployment = checkout` was labelled "the alert's name says so": the name only chose which label to prefer, and the chip pointed at a line the value is not on.

**The Open button went grey with no reason.** Two kinds this app recognises in a label have no page here, and a namespaced kind whose alert named no namespace cannot build a URL either. Both say which now, beside the button.

**`[FIRING:5]` was folded into one reading silently.** Five alerts, five sets of labels, one of them read — the panel says how many the message carried and that the fields are the first one's.

**`?since=` went where nothing reads it.** Only the Changes tab reads it, and only three kinds have one, so the link now opens that tab on those kinds and stays plain otherwise; `ALERT_WINDOW_MS` was exported and used nowhere. Four more rules that survived deletion with the suite green got tests.

PR #160, opened 2026-09-10
fix(logs): a refusal the kubelet returned with a 200 is not a log line

belliel:
Found by pointing the ignored `live_*` harnesses at a real v1.36 cluster, which is the first time that has been done here.

## What is wrong today

Ask for a container's previous run on a node whose container runtime has already collected it, and the kubelet answers **HTTP 200, `text/plain`**, with this as the entire body:

```
unable to retrieve container logs for containerd://3bb6fd00e2a1...
```

No status code to read, no `Status` object, nothing structurally different from output. `get_logs` hands the body to `parse_logs`, so the log panel draws it as a line the container printed. Through `useHintChain` it also becomes an input to "Most likely" and a quoted line in "Copy for agent": a diagnosis built on a sentence no program ever wrote.

The contrast is what makes it clear. Asked for the previous run of a container that has **never restarted**, the same apiserver answers properly:

```
$ curl .../log?container=seed&previous=true
{"kind":"Status","status":"Failure","message":"previous terminated container \"seed\" ... not found","code":400}
```

400 and a `Status`, which the code already recognises as `NoPreviousRun`. The other one is a failed read answered with content, which is this repo's one defect class aimed at the log panel.

## What changes

A third answer, because it is a third thing. `NoPreviousRun` is "there was never anything to read". `Broken` is "the transport failed and the thing may still be running, so offer a retry". This is neither: the container restarted, so the run happened, and the node dropped its log before anyone asked. Retrying reaches the same node, which still does not have it.

- `Error::LogNotKept { container, said }`, carrying the kubelet's own words rather than replacing them.
- `StreamFailureKind::LogNotKept`, arriving as `log-not-kept`. Adding it made the compiler point at `terminal/manager.rs`, the second reader of that map, exactly as intended.
- `is_runtime_dropped_log`, checked in `get_logs` before parsing and in the follow loop's failure arm.
- The panel says the node no longer has that log and offers the current run, in warn rather than err, with no reconnect: there is nothing to reconnect to. The legend chip reads `· log not kept`.

**Matched as the whole body, not as a substring.** A program that prints that sentence among its own lines keeps its logs, and hiding them behind a refusal would be the same lie pointed the other way. A test holds both directions.

## How to check it

- `cargo test --workspace` — three new tests in `state/events.rs`, over the exact body recorded from the cluster: the refusal is recognised, a container printing it among other lines is not, and the neighbouring 400 still classifies as `NoPreviousRun`.
- Against a live cluster: `KUBECONFIG=... cargo test --test live_init_containers -- --ignored --nocapture`. Before the fix, on a node that had collected the previous container, it printed `migrate/previous: unable to retrieve container logs for containerd://...` and failed on the assertion that the previous run is the one that printed the error.

I cannot make the node drop a log on demand, so this is not reproducible at will; the body in the test is the one the cluster actually returned, captured with the status code and content type.

## Risk and rollback

Small and one-directional. Before, this body always became a log line; now it becomes a named failure, and nothing else changes shape. `terminal/manager.rs` folds the new kind into the arm it already used for the kinds an exec cannot produce. Reverting restores the old behaviour.


belliel:
Confirmed live on a fresh v1.36.1 sandbox (2026-09-11), and the sandbox produced the case on its own.

The `init-demo` specimen's `migrate` init container had restarted five times; asking for its previous run got a 200 with this body and nothing else:

```
unable to retrieve container logs for containerd://50dc789fec2c62497081e1f02fe86eaeec194d6052515814dd341a1519a70cfa
```

On main that is one log line in the pane. On this branch `get_logs` answers `Error::LogNotKept { container: "migrate", said }`, classified `LogNotKept`; `wait-for-db` next to it still answers `NoPreviousRun`, and a missing pod still `Gone`.

The live harness (`live_init_containers.rs`) used to demand the previous run's output and failed here, which was the harness asserting something the node had not kept. It now accepts either honest answer, the printed error or "not kept", and rejects the third: the kubelet's sentence as a line the container wrote.

```
migrate/previous: not kept by the node: unable to retrieve container logs for containerd://50dc789f…
wait-for-db/previous: No previous run: wait-for-db has not restarted, so there is no earlier log
test init_containers_and_previous_runs_reach_the_frontend ... ok
```

Dudude-bit:
Ran the review workflow over the branch (5 dimensions, each finding put to independent refutation): 20 survived, 9 distinct. Fixed in `9bf5668`.

**The refusal was caught on the wrong path.** `is_runtime_dropped_log` had exactly one call site — `get_logs`, the one-shot behind Download. The panel streams: `useLogStream` → `stream_pod_logs` → `stream_logs`, where the body went straight into `parse_log_line`. So the pane still drew the kubelet's sentence as a line the container printed, `StreamFailureKind::LogNotKept` was never emitted by anything, and every branch this PR added — `notKept` in LogViewer, the legend chip, both catalogue strings — was unreachable. Download reported it correctly while the pane beside it lied about the same fact.

The stream now holds the first line back when it *is* that sentence, releases it the moment anything follows (or after one 50 ms flush interval), and reports `LogNotKept` if the stream ends with it still held — the whole-body rule, applied to a stream. It lives in a `FirstLine` struct with four tests, three of which fail when the hold is deleted.

**A 400 carrying the same sentence classified as `Broken`** and got a Reconnect that reaches the same node for ever. `classify` recognised `LogNotKept` only structurally, while its sibling has both a structural check and a text fallback for exactly this reason. Added `mentions_dropped_log` — a substring rule, used only on error text, never on a body.

**`stranded` did not count it**, so `LogHistoryBar` returned `null` on the one read a history vendor could still answer: the log is nowhere on the apiserver, and Loki is the only thing left.

**Download aborted the rest.** One `try` around the whole container loop meant the first dropped log took every container after it, and the toast said the API could not be read — over a fact the backend had just named exactly. Per-container now, naming which ones and why.

**The copy asserted a restart nothing read.** "It restarted, so the run happened" — the runtime drops a *running* container's log too (containerd restarted, disk pressure, the container pruned), and `is_runtime_dropped_log` never consults `previous`. It says only what was read now, and "Show the current run" is offered only when the failed read was the earlier one.

Smaller: `· log not kept` was a raw English literal — as were its four siblings, so all five chips are catalogue strings now; the kebab-case guard did not list the new kind; and the `!body.contains('\n')` clause — the half the doc comment is entirely about — had no test, so deleting it left the suite green. The case it exists for (a wrapper whose *first* line is that sentence and which keeps going) is in now.

Gates: tsc, lint, 2584 frontend tests, 613 Rust tests, clippy, comment budget 15%.

Dudude-bit:
Checked live on a kind cluster with the app open: the current run streams unaffected across five containers, and the previous run reads 8 000 lines with no false notice — the held first line costs nothing on an ordinary stream.

The case that *is* manufacturable is the one pointed the other way, and it holds: a pod whose only output is exactly that sentence has its line drawn as a log line with its timestamp, not as a refusal. The apiserver's RFC3339 prefix is why — a line that came from the container never starts with the sentence.

The refusal itself I still cannot make a node produce on demand; it is covered by the four `FirstLine` tests and the panel test, each verified by deleting the branch and watching the right test fail.

PR #161, opened 2026-09-10
fix(connections): a claim list the cluster refused is not a missing claim

belliel:
Found by pointing the app at a real cluster with an identity narrow enough to be refused. Independent of the open stack; based on `main`.

## What is wrong today

This is the sentence [CLAUDE.md](../blob/main/CLAUDE.md) opens with, and it is in `connections.rs`:

> `Vec::new()` on a 403 becomes "there are none" and then "this Gateway does not exist", in red, with nothing failing anywhere.

Here it is a claim rather than a Gateway. `Snapshot::of` read the namespace's claims as

```rust
claims: claims.map(|list| list.items).unwrap_or_default(),
```

so a token without `persistentvolumeclaims` arrived at `named_object` with an empty list, and every claim a pod mounts fell to the `None` arm:

```rust
None => ObjectRef::new(kind, name, Some(ns.to_string()), Existence::Missing),
```

A person whose role does not include PVCs opens a perfectly healthy pod and is told its volume **does not exist**. Nothing fails, nothing is logged, and the reader is sent hunting a storage outage that is not happening.

The function's own doc comment shows how it happened. It says claims "were listed, so they carry their phase and size and can be called present or missing" — sound reasoning, resting on a premise that `unwrap_or_default()` had quietly stopped guaranteeing.

Recorded from the cluster, before:

```
claim edges: 1
  - pvc-demo existence=Missing
```

and after:

```
not_looked_at: 3 kinds
  - PersistentVolumeClaim : Unanswered { version: "v1", said: "...persistentvolumeclaims is forbidden..." }
claim edges: 1
  - pvc-demo existence=NotChecked
```

## What changes

`Snapshot.claims` becomes `Read<PersistentVolumeClaim>`, which is the type the four fields beside it already use for exactly this reason. Three consequences, and no others:

- `named_object` returns `notChecked` for a claim whose list was refused, joining the `ConfigMap`, `Secret` and `ServiceAccount` references that have always said that. A list that really answered and really lacks the claim still says `missing`, because that one is the app finding a broken pod and worth saying.
- `unanswered()` names the refused claim list, so the reader gets the explanation in the one place the page collects them.
- Nothing else: the compiler found no other reader.

## How to check it

- `cargo test --workspace` — three tests in `commands/connections.rs` over the words the apiserver actually said. Deleting the `Err` arm of `named_object` fails the first; the second holds the opposite direction so the fix cannot be "always say notChecked".
- `tests/live_refusals.rs` is new: an ignored harness that points the app at a deliberately narrow identity and asserts the neighbourhood names what it could not read. Its header has the four commands that build the identity. Run:

```
K8S_GUI_REFUSED_CONTEXT=narrow K8S_GUI_REFUSED_POD=shell-demo \
  cargo test --test live_refusals -- --ignored --nocapture
```

## A second defect, in the same struct, not fixed here

The same live run found this, and it is a different shape and a bigger change, so it is not in this PR:

```rust
pods: pods?.items,
services: services?.items,
ingresses: ingresses?.items,
```

An identity that can read pods but not Services gets **no neighbourhood at all**: `connections_of` returns the raw 403 about `services`, so the Connections tab dies whole even though the subject pod is perfectly readable. Loud rather than silent, so less dangerous than the claim, but the pod is right there and the answer should be the pod plus the names of what was refused. Happy to take it next if you want it.

## Risk and rollback

Contained. The only behaviour that changes is a claim whose list was refused, which went from `missing` to `notChecked`, plus one more entry in `not_looked_at`. Every path where the list answered is untouched, and a test holds that. Reverting restores the old struct field type.

### DIFF
--- a/src-tauri/src/commands/connections.rs
+++ b/src-tauri/src/commands/connections.rs
@@ -704,7 +704,7 @@ fn uses_from_spec(
     ns: &str,
     subject: &ObjectRef,
     spec: &PodSpec,
-    claims: &[PersistentVolumeClaim],
+    claims: &Read<PersistentVolumeClaim>,
     out: &mut Neighbourhood,
 ) {
     let mut targets: Vec<(String, String)> = Vec::new();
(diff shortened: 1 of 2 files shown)


Dudude-bit:
Reviewed with current `main` merged in (the branch was 35 commits behind; clean). The change itself is right and narrow: a refused claim list becomes `NotChecked` rather than "missing", and gets named in `not_looked_at` with the cluster's own words.

**Its live harness could not fail, three ways.** Run exactly as the file's own doc comment prescribes:

- The role it tells you to create allows pods only — and `Snapshot::of` takes pods, services and ingresses with `?`. Under that identity `connections_of` returns `Err` and the file panics on `.expect("the pod itself is readable")` before a single assertion runs. The recipe now allows those three and refuses the claim list, which is the read this file is about.
- The default subject was `log-demo`, which is a Deployment and no pod at all, so the read was `not found` even on a wide identity.
- The one assertion about the fix walks the claim edges, and a pod that mounts no claim left it walking nothing. It refuses to pass without one now.

Verified against a real 403 on kind — service account, role, rolebinding, impersonating context:

```
subject: "Pod" shell-demo
not_looked_at: 3 kinds
  - PersistentVolumeClaim : Unanswered { …persistentvolumeclaims is forbidden… }
claim edges: 1
  - pvc-demo existence=NotChecked
```

and with the fix reverted it fails with *"pvc-demo was drawn as Missing, which tells a person a mounted, healthy volume does not exist"*. That is the file doing what it promised.

Pushed onto the branch.

PR #162, opened 2026-09-10
fix(connections): a refused list is not an empty one, and not a missing object

belliel:
Stacked on #161, which found the first half of this in the same struct. Review the last commit.

## What is wrong today

`Snapshot::of` took ten lists and treated three of them as load-bearing:

```rust
pods: pods?.items,
services: services?.items,
ingresses: ingresses?.items,
```

**One refusal killed the whole answer.** An identity allowed to read pods but not Services got no neighbourhood at all: `connections_of` returned the raw 403 about `services`, and the Connections tab died whole while the subject pod sat there perfectly readable. A neighbourhood is a set of independent reads; one the cluster refused is a gap in the answer, not the end of it.

**And the subject was looked up inside those same lists**, so a refused list reported the object itself as gone:

```rust
let svc = snapshot.services.iter()
    .find(|svc| svc.name_any() == name)
    .ok_or_else(|| Error::not_found("Service", name, ns))?;
```

A token without `services` opens a Service and is told it **does not exist**. That is the third instance of the same defect this pair of PRs is about, and the most expensive one: it sends a person to rebuild something that is running.

## What changes

The three lists become `Read<T>`, the type the four fields beside them already use.

- **Nothing aborts.** Each refused list is named in `not_looked_at`, alongside the autoscalers and budgets that have always carried theirs, so a group that goes quiet has a reason next to it instead of reading as "there are none".
- **`found()` keeps the two answers apart**, and it is the only place that does. A list that answered and does not hold the name still means the object is gone, and says so. A list nobody could read means nobody knows, and returns the new `Error::ListUnread` carrying the cluster's own words. Deliberately not `Error::NotFound`, which asserts.
- **Accessors, not raw fields.** `snapshot.pods()` returns the items or none, with a doc comment naming the one rule: a caller turning an empty answer into a *statement* cannot lean on it alone.
- **The frontend's one affirmative negative goes quiet with it.** `chainSilence` said "No Service in this namespace selects this pod", which with Services unread is the app reporting its own blind spot as a fact about the cluster. It now returns null when `Service` is among the unread kinds, and the "Not looked at" group carries the explanation instead.

## How to check it

- `cargo test --workspace` — four tests in `refused_list_tests`: a refused list makes the subject unknown rather than gone, a list that answered still reports a missing object as missing, the happy path, and every refused list reaching `not_looked_at`.
- `bun run test` — `connections.test.ts` holds both directions of `chainSilence`.
- Live, with the narrow identity from #161's harness header:

```
kubectl patch role pods-only -n k8s-gui-test --type=json \
  -p '[{"op":"replace","path":"/rules","value":[{"apiGroups":[""],"resources":["pods"],"verbs":["get","list","watch"]}]}]'
K8S_GUI_REFUSED_CONTEXT=narrow cargo test --test live_refusals -- --ignored --nocapture
```

Before, that run ends in a 403 about services and no answer at all. After, the pod comes back with `not_looked_at` naming Service, Ingress, PersistentVolumeClaim, HorizontalPodAutoscaler and PodDisruptionBudget.

The final live confirmation is outstanding: the sandbox's `kubectl proxy` died mid-run, and the run I got back is the one where `found()` answered a 502 with `ListUnread { kind: "Pod" }` instead of "not found", which is the same distinction from the transport side. Everything else here is covered by the unit tests.

## Risk and rollback

The behaviour that changes is all in one direction: reads that used to abort now continue, and objects that used to be called missing are now called unknown. Nothing that answered behaves differently, and tests hold that from both sides. `Error::ListUnread` is new, so nothing matches on it yet; the frontend classifies it as a refusal through the message, which carries the apiserver's "is forbidden".

### DIFF
--- a/src-tauri/src/commands/connections.rs
+++ b/src-tauri/src/commands/connections.rs
@@ -249,7 +249,7 @@ fn traffic_into(
     snapshot: &Snapshot,
     out: &mut Neighbourhood,
 ) {
-    for svc in &snapshot.services {
+    for svc in snapshot.services() {
         let selector = service_selector(svc);
         let Some(text) = Selector::Equality(&selector).says() else {
             continue;
@@ -453,7 +453,7 @@ fn gateway_traffic_into(
 
 /// The Ingresses whose backend names this Service.
 fn routes_into(ns: &str, svc_ref: &ObjectRef, snapshot: &Snapshot, out: &mut Neighbourhood) {
-    for ing in &snapshot.ingresses {
+    for ing in snapshot.ingresses() {
         for (backend, relation) in ingress_backends(ing) {
             if backend != Backend::Service(svc_ref.name.clone()) {
                 continue;
@@ -496,7 +496,7 @@ fn note_reach(
     let selector = service_selector(svc);
     let query = Selector::Equality(&selector);
     let selected: Vec<&Pod> = snapshot
-        .pods
+        .pods()
         .iter()
         .filter(|pod| query.matches(pod.labels()))
         .collect();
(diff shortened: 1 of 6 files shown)


belliel:
Live confirmation, which the body said was outstanding. Fresh v1.36 sandbox, test manifests applied, the `narrow` identity from #161's harness header with a Role of `pods` only:

```
K8S_GUI_REFUSED_CONTEXT=narrow K8S_GUI_REFUSED_POD=shell-demo \
  cargo test --test live_refusals -- --ignored --nocapture

subject: "Pod" shell-demo
edges: 4
not_looked_at: 5 kinds
  - HorizontalPodAutoscaler : Unanswered { version: "autoscaling/v2", said: "…is forbidden…" }
  - PodDisruptionBudget : Unanswered { version: "policy/v1", said: "…is forbidden…" }
  - PersistentVolumeClaim : Unanswered { version: "v1", said: "…is forbidden…" }
  - Service : Unanswered { version: "v1", said: "…is forbidden…" }
  - Ingress : Unanswered { version: "networking.k8s.io/v1", said: "…is forbidden…" }
claim edges: 1
  - pvc-demo existence=NotChecked
test result: ok. 1 passed
```

Before this branch the same run ended in the raw 403 about `services` and no neighbourhood at all. The pod comes back, every refused list is named, and the claim is `NotChecked` rather than `Missing`.

PR #166, opened 2026-09-11
feat(keys): one table of every shortcut, drawn behind ?, and a test that keeps it complete

belliel:
Small item from the master plan ("Keyboard: overlay `?`, `g`+letter, `l`, `y`; single source for shortcuts; completeness test"). Based on `main`.

## What is wrong today

Shortcuts live in six `keydown` listeners spread across the tree (the palette's `mod+k`, Settings' `mod+,`, the tab strip's `ctrl+tab`/`mod+t`/`mod+w`/`mod+1..9`, the copy-link `mod+shift+c`, the log viewer's digits, the builder's `del`/`mod+a`/`mod+shift+i`) and nowhere else. There is no list a person can open, and nothing notices when a new one is added. The plan asks for a single source and a test that it is complete, because a list that is complete today and quietly stops being complete is worse than no list.

## What changes

**One table.** `src/lib/shortcuts.ts` names every shortcut with its keys, its section and its sentence. The overlay behind `?` draws it; the central handler reads its chords from it; nothing else defines a key.

**The completeness test.** `shortcuts.test.ts` walks `src/` for `addEventListener("keydown"` and refuses any file the table's `KEYDOWN_SITES` does not name, and refuses a named file that no longer listens. Sabotaged before opening this: a scratch file with one listener fails it with the file's path in the message. A second test holds that every entry has a sentence in both languages, a third that no two entries in a section share keys, a fourth that every chord goes somewhere and every page key names a tab.

**Two things the app did not have.** `g` then a letter goes to a list: `g p` pods, `g d` deployments, `g s` services, `g i` ingresses, `g n` nodes, `g e` events, `g j` jobs, `g c` configmaps, `g o` overview. On an object's page a letter opens a tab: `o` overview, `l` logs, `y` YAML, `e` events. The page keys go through `ResourceDetailLayout`, the one frame every kind renders through, so a page without that tab simply does not move and nothing has to be wired per page.

**What stays quiet.** Unmodified keys only, never from a field, a terminal or an open Radix layer: those own their keys, and a `g` typed into a search box is a letter. A chord that waits longer than 1.5 s is two letters, not one shortcut.

## How to check it

- `bun run test`: `shortcuts.test.ts` (4), `useShortcuts.test.tsx` (5: chord navigates, a stale `g` is forgotten, `?` toggles, a field is left alone, a page key asks the layout), `ShortcutsOverlay.test.tsx` (2: every table entry is on the overlay, a chord reads "g then p").
- By hand: `?` anywhere; `g p`; on a pod, `l`.

Rendered from source only, not opened in the app.

## Risk and rollback

Additive. Every existing listener is untouched and merely named in the table. The one new global listener ignores modified keys and anything inside a field or a layer, so it cannot shadow the palette, Settings, the tab strip, the terminal or the builder. Reverting removes the table, the handler, the overlay and the four page keys.


Dudude-bit:
Reviewed with current `main` merged in (30 commits behind; clean). Four fixes pushed, plus one the merge itself needed.

**Your own completeness guard caught the merge.** `#204` landed a `keydown` listener in `ClusterList` while this was open, and the guard failed on it the moment main came in. Named in `KEYDOWN_SITES` with the condition it is claimed under — only on the screen that *is* the cluster list — and listed in the overlay. Nice to see a guard earn its keep on day one.

**The overlay spelled shortcuts itself.** `keyWord()` rendered `mod+k` as "⌘+k" while `ScopeTabs` — always on screen — drew the same key as "⌘K" through `formatShortcut`. On Windows it produced "Ctrl+⇧+c", a ⇧ glyph where the app deliberately writes "Shift"; on macOS "Ctrl+Tab" where the rule is ⌃. It also read `navigator.platform`, deprecated, while the app takes the host OS from the backend. It uses `formatShortcut` now, and the named keys (`esc`, `enter`, `tab`, `del`, `space`) moved into that one renderer — spelled as the app already spelled them, so no other screen changes.

**`?` opened the list and could not close it.** The list is itself a Radix dialog, and the handler stepped aside for any dialog in the document. The test that said otherwise renders only `Probe`, never the overlay — so it passed against behaviour the app does not have. There is a test that mounts it now.

**That whole-document test was wrong in both directions**, which is the substantive one. The peek panel is a non-modal Sheet, opened that way on purpose so the reader keeps working behind it, and a plain row click opens one — so every unmodified key was silently dead for as long as a peek was up, while ⌘K and ⌘T kept working. And a dropdown carries `role="menu"` with no dialog at all, so its keys were fighting the global ones. The question is *where the focus is*, which `ClusterList` had already settled and documented as learned twice. One owner now, `claimedByLayer`, beside `claimedByTarget`.

Checked in the app: `?` opens and `?` closes; with a peek open and focus in it the chords stay quiet, and after clicking back into the list behind it `g d` navigates again.

**The table promised every key and omitted five** the app prints on screen: `mod+s` and Backspace in Files (the Files tab prints its own hint for `mod+s` two screens from an overlay that had never heard of it), `mod+a` in the log viewer, Enter on the front door, and the palette's own `mod+Enter`. Added, with a second guard: every shortcut the app draws a hint for has to be in the table. The keydown sweep cannot see these — they are React `onKeyDown` props — and that is exactly the gap they fell through.

Dudude-bit:
Second pass, two more fixes pushed.

**Caps Lock.** `event.key` for a chord typed with Caps Lock on arrives uppercase, so every single-character shortcut stopped answering. Normalised where the key is read, leaving named keys alone.

**The completeness guard only saw `addEventListener("keydown")`.** A shortcut written as a React `onKeyDown` prop was invisible to it, and the second guard could not catch those either, because it only knows shortcuts the app draws a hint for. Three files answer ⌘-chords that neither guard could see: `ScopeTabs` (mod+click / mod+Enter add a namespace instead of replacing it), `LogList` (mod+a selects the rendered window rather than the whole buffer) and `FilesTab` (mod+s downloads). The sweep now also matches an `onKeyDown` file that reads a modifier — a handler that only looks at `event.key` is a control answering its own Enter, not a shortcut — and all three are named in `KEYDOWN_SITES`. Dropping any one of them fails the test.

PR #167, opened 2026-09-11
feat(checks): test a hypothesis from the pod, without kubectl

belliel:
The last small item from the master plan ("Hypothesis without kubectl: DNS, TCP, pod file; marked as copy; guaranteed cleanup"). Based on `main`.

## What is wrong today

The app can say *most likely it cannot reach its database*, and then the person opens a terminal. A DNS or TCP question has to be asked from the pod's own network to mean anything, and the only way to ask was `kubectl exec` or `kubectl debug` by hand, with the answer read off a shell and the debug pod left behind.

## What changes

A **Checks** tab on the pod page with two questions: *does this name resolve*, *does this host:port answer*. Both are asked from where the pod stands, never from this machine.

**In the pod's own container first.** A ladder of tools, most common first: `getent`, `nslookup`, `host` for names; `nc`, `curl` for ports. A rung the image does not have (exit 127, or the runtime's own "executable not found") moves to the next; when none is there the answer says so and names what was tried, which is a fact about the image and not a failed check.

**Every rung is argv.** The name or host a person typed reaches the container as one argument and nothing else. `bash -c 'exec 3<>/dev/tcp/…'` would reach more images and is deliberately not on the ladder: it needs the host in a shell string. A test types `db; rm -rf /` and asserts it is one argument in every rung and refused by validation before that.

**From a copy when the image has nothing.** The same exec into a throwaway pod: the original's namespace, labels, DNS policy, DNS config, service account, host-network flag, tolerations and pull secrets, and none of its containers, volumes or owner references; one container that runs `sleep` and waits to be asked. It carries `k8s-gui/check-pod=true`, `restartPolicy: Never`, `activeDeadlineSeconds: 300` and a zero grace period. The answer is marked as coming from a copy, with the image and the pod name, because a copy is not the pod.

**Guaranteed cleanup.** `CopyGuard` deletes the copy on `Drop`, which a cancelled future, an early `?` and a panic all run; it spawns the delete because a destructor cannot await. The happy path deletes explicitly so the report can say *deleted afterwards* rather than *asked to delete*, and the frontend shows *not confirmed deleted* when it could not. `activeDeadlineSeconds` is the last line: a copy nothing ever came back for still leaves on its own.

**Words at render, from facts.** The backend returns which tool answered, its output, its exit and where it ran. busybox `nslookup` exits 0 with "can't resolve" in its output, so a zero exit is not an answer: an address in the output is, and the resolver's own address in the `Server:` header is not one. A test holds both.

## How to check it

- `cargo test --workspace`: `commands::checks` (5): a typed host is one argv argument and is refused before that; a host may be a name or an IP literal, a port may not be 0; the copy keeps the network identity and drops everything else; a copy name stays a label however long the pod name; an image with no tool is said so and not called a failure.
- `bun run test`: `checks.test.ts` (5, including the nslookup zero-exit case), `ChecksTab.test.tsx` (3: the pod's own container is asked first with what was typed; a copy is offered when the image has no tool and the answer says it came from one; no port, no run).
- By hand on any pod: Checks tab, type `kubernetes.default.svc.cluster.local`, Run; type `kubernetes.default.svc.cluster.local:443`, Run. On a distroless image, the answer offers *Run from a copy*.

### Run against a live cluster, and what it found

`tests/live_checks.rs` (added in the second commit) runs `check_pod` against a real v1.36 cluster with the test manifests applied:

```
K8S_GUI_CHECK_POD=meshed-demo-… K8S_GUI_CHECK_CONTAINER=app \
K8S_GUI_BARE_POD=pebble-… K8S_GUI_BARE_CONTAINER=pebble \
  cargo test --test live_checks -- --ignored --nocapture --test-threads=1

dns:  ran_in: container, tried: [getent, nslookup], answered_with: nslookup, ok: true   (353 ms)
tcp:  ran_in: container, tried: [nc], answered_with: nc, ok: true                        (233 ms)
bare: ran_in: container, tried: [getent, nslookup, host], answered_with: None, tool_missing: true
copy: ran_in: copy, answered_with: nslookup, ok: true, copy: { image: busybox:1.36, deleted: true }  (1964 ms)
test result: ok. 2 passed
```

The busybox pod has no `getent`, so the ladder moved to `nslookup`, which is the fall-through working. The pebble image has none of the three, which is said so. The copy answered and reported deleted; `kubectl get pods -l k8s-gui/check-pod=true` is empty afterwards.

**The first live run of the copy failed, and the failure is the third commit.** The copy carried the original's labels and had no owner, so the original's ReplicaSet adopted the label-alike orphan and deleted it as surplus within a second, before the first `get`:

```
SuccessfulDelete  replicaset/pebble-6cf79cb988  Deleted pod: pebble-6cf79cb988-kh8mj-check-1789120025
```

Two lines fix it, and a unit test holds each. The copy now carries a **controller owner reference to the pod it copies**: a pod with a controller is never adopted, and if the original goes the garbage collector takes the copy with it. And because the same labels are what a Service selects, and a pod with no readiness probe is Ready the moment it runs, the copy carries a **readiness gate nothing ever satisfies**: never Ready, so never an endpoint of anything while it lives. The gate is by construction, not observed: the copy lives about two seconds, and the pod-level Ready condition is what EndpointSlices read.

The copy path could not have been found without a cluster: nothing in a unit test adopts a pod.

## Risk and rollback

Nothing runs unless a person presses Run. In the pod's own container the exec is one short read-only command with a 20 s timeout. The copy path creates one pod, labelled, with a 300 s deadline, and deletes it. Reverting removes the tab, the command and the module; nothing else changes shape.

Out of this PR, on purpose: the "pod file" hypothesis from the plan item. The Files tab (#134, merged) already reads a file's existence and content from the container with the same exec-without-TTY machinery and the same debug-container path, so a third form here would have been the same read behind a second door.


Dudude-bit:
Ran the review workflow over the branch (4 dimensions, each finding put to independent refutation). What survived is fixed on this branch; the live check on a kind cluster found one more.

**A port that answers is not a port that refuses.** `curl telnet://` exits about the transfer, not about the port: it connects, waits for bytes an ordinary TCP service never sends, and `-m 3` kills it with 28. Measured against a local listener that accepts and stays silent — 28 for an open port, 7 for a refused one. Reading 0 as the only yes made the fallback rung report every healthy service as "does not answer from here". Each rung now states what its own exit means.

**A check that produced no answer is not a check that said no.** `Exit` carries three states and `CheckOutcome` kept two, so an exec whose status channel never produced one — a dropped websocket, an apiserver that went away mid-exec — rendered as "the name does not resolve": a stated fact about the cluster from a run that produced no fact at all. Both the outcome and the verdict carry the third state now.

**The copy could not be admitted where it is most needed.** A namespace enforcing restricted PodSecurity rejects a pod that states no `securityContext` — exactly the namespace with no `kubectl debug` to fall back to. And a hostname was validated as a Kubernetes resource name, so `Payments.example.com` and a trailing-dot FQDN were refused outright.

**The guard was built after the create**, so an accepted pod whose answer never came back left a copy running that nothing in this process knew about, with `activeDeadlineSeconds` as the only thing behind it. Deleting a copy that was never created is a 404 and costs nothing. And "this pod is a throwaway this app made" is one fact with two readers, both keyed on the debug label alone: the delete command refused a check copy as "not created by k8s-gui", and the pod page never offered to remove it.

**"Can this container take an exec" is one judgement**, with a doc comment saying so and three surfaces reading it. This tab was a fourth that did not: it offered a terminated container as though it could answer, and the run failed with the runtime's words instead of ours. A source guard now holds every offering surface to asking, because the chooser is a Radix Select whose popup does not open under jsdom.

**The test that could not fail.** Six fields the copy must not inherit were asserted against a fixture that never set them — adding `node_name` back to `copy_of` passed. The fixture carries a node, a node selector, an init container, an affinity and a status now, and each assertion fails when the field survives.

**Found with the app open, not by reading:** `runAsNonRoot` with no `runAsUser` is `CreateContainerConfigError` on every image whose default user is root — busybox among them, which is the image the panel offers. My own hardening broke the feature: the copy never started, and the reader was told the check could not run. 65534 satisfies restricted PodSecurity and runs `sleep`, `nc` and `curl`.

PR #168, opened 2026-09-11
fix(yaml): a long manifest scrolls again (#163)

### LINKED ISSUE #163: Отсутствует скрол для просмотра YAML
Версия 4.11.0
Windows 11

При открытии большого YAML (из ConfigMap, Deployment, etc) в окне нету скрола (вертикального) и не скролируется по mouse wheel

Dudude-bit:
Closes #163.

## What the reader saw

A ConfigMap or Deployment with a long manifest opened with no vertical
scrollbar, and the mouse wheel did nothing. Reported on 4.11.0 / Windows 11;
reproduced here on a 412-line ConfigMap.

## Why

`@uiw/react-codemirror` mounts CodeMirror inside a wrapper `div` of its own,
and the `height` prop only becomes a theme rule on the `.cm-editor` beneath
it. A percentage there resolves against that wrapper — which had no height, so
it collapsed to `auto`. The editor then grew to the height of the whole
document, the ancestor `overflow-hidden` clipped it, and there was nothing left
to scroll.

Two of the three surfaces had it: the detail page's YAML tab and the peek
panel's, neither of which passes a `className`. The edit dialog escaped only
because it happens to pass `className="h-full"`, which put a height on that
same wrapper by accident. That is why the fix is in the component and not at
each call site — the next surface would have had to remember too.

## Also

`.cm-scroller` is built by CodeMirror, so the app's `scrollbar-thin` never
reached it. On a platform that lays scrollbars out rather than overlaying them
the manifest was the one pane with a platform-default bar down its side; it now
carries the same thin one as every other scroller here.

## How it was checked

Live, on a 412-line ConfigMap, with real wheel events over each surface:
before, the peek panel stayed on line 1 after eight wheel-downs; after, it
reaches line 46, the detail tab line 129 and the dialog line 105. The Data tab
next to it scrolled the whole time, which is how I knew the harness was sound.

Unit: `YamlEditorImpl.test.tsx` asserts the wrapper is handed the height (all
three call shapes), and `editor-theme.test.ts` asserts the scrollbar rules and
that they use role tokens. Both fail when the fix is removed — checked.

### DIFF
--- a/src/components/resources/CustomResourceList.tsx
+++ b/src/components/resources/CustomResourceList.tsx
@@ -351,8 +351,13 @@ function formatColumnValue(
       if (typeof value === "string" && isKnownStatus(value)) {
         return <StatusBadge status={value} />;
       }
+      // A printer column is whatever the CRD author chose to show; some are
+      // long, and the cell is the only place it appears.
       return (
-        <span className="max-w-[200px] truncate text-fg-mid">
+        <span
+          className="max-w-[200px] truncate text-fg-mid"
+          title={String(value)}
+        >
           {String(value)}
         </span>
       );
--- a/src/components/resources/GatewayRoutesList.tsx
+++ b/src/components/resources/GatewayRoutesList.tsx
@@ -508,7 +508,13 @@ export function GatewayRoutesList() {
       ))}
 
       {showMap && (
-        <div className="mt-3 flex flex-col gap-1">
+        // Bounded and shrink-proof. The map's board is as tall as the spine
+        // it draws — 56px a node — and as a full-height sibling of the only
+        // scroller it took every pixel: on a cluster with a dozen or more
+        // routes, showing the map left the route list below it with a box
+        // 0px tall, so the rows the header's own counts link to were not on
+        // screen at all and there was nothing to scroll.
+        <div className="mt-3 flex max-h-[45%] shrink-0 flex-col gap-1 overflow-y-auto scrollbar-thin">
           {topology.columns[topology.spine ?? 1].nodes.length === 0 ? (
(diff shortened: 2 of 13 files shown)


Dudude-bit:
Pushed a second commit from an audit for the same shape elsewhere — a box
whose height comes from a percentage or `h-full` where an ancestor has none,
so the child grows past its container and something above clips it.

**Dialogs had it, and worse.** A panel is centred with `translate-y-[-50%]`
and had no ceiling, so an unbounded body grew off the top *and* the bottom of
the window — the title and the ✕ off one edge, the buttons off the other, with
nothing to scroll: Radix locks the page behind the panel, and a fixed box
cannot be scrolled into view by an ancestor. The drain dialog lists every pod
still on the node, so a drain stalled on a PodDisruptionBudget could not be
stopped from the dialog that started it. Three call sites had each bounded
themselves by hand; the rest had not.

Also: the Routes page's map was a full-height sibling of the only scroller, so
showing it left the route list a box 0px tall; `kubectl apply`'s output in the
YAML dialog was an uncapped `<pre>` that squeezed the editor and pushed the
footer off; and that block said "Success" and "Error" in English.

**What I saw versus what I reasoned.** The YAML scroll, the dialog fix's
absence of regressions, and the drain confirm at the 768px minimum window I
checked on screen. The routes-list collapse I did not reproduce — this cluster
has Istio Gateways rather than Gateway API, so there was no map to draw; that
one rests on a measured CSS replica and is a defensive bound plus a floor.

Two findings I left alone as a different defect, not this one: an Event
message and a CRD printer-column value are truncated with no tooltip and no
other way to read them in full. Worth their own issue if you agree.

PR #169, opened 2026-09-11
feat(yaml): before applying, ask the server what would change

belliel:
The follow-up the Trust package left open ("server dry-run before apply", from Critical Infrastructure, #129). Based on `main`.

## What is wrong today

The apply confirmation shows a diff of the buffer against the file the editor was given. That is what the person typed. It is not what the cluster will do: the server fills defaults, admission mutates, and a validating webhook may refuse the whole thing. None of that is visible until after Apply, which on a cluster marked critical is exactly the moment it should have been visible before.

## What changes

**The server is asked first.** The same server-side apply `apply_manifest` runs, with `dryRun=All`. The apiserver runs admission, fills defaults and answers with the object it would have stored, and stores nothing. The confirmation shows that object against the one the cluster holds now, both cleaned the way the editor shows them (`get_manifest`'s path: private keys redacted, bookkeeping stripped), so a Secret stays a Secret on both sides of the diff.

**Six answers per document, because the empty cases mean opposite things.**

- *would be created*: no object there, the server accepts it.
- *would change* / *would not change*: the server's object against the current one.
- *accepted, but the current object could not be read*: a 403 on the `get` with a successful dry run. Drawn as "would be created" this tells a person they are about to make something that already exists; it gets its own sentence and a unit test that fails if it ever collapses into `Created`.
- *refused*: a 4xx with a `Status`, the server's own words shown. **Apply is disabled**, because a real apply would be refused the same way.
- *unanswered*: a 5xx, a transport failure, a proxy in the way. Found by the live run: a sandbox proxy's 502 was landing in `Refused` and would have blocked an apply the server never saw. Not a refusal, so it does not block; the editor's own diff stands in and says so.

**Nothing is lost while the answer is on its way.** The editor's diff shows with "asking the cluster what would change…", and if the dry run itself fails, with the failure in words. The critical-cluster typed-name gate is untouched and still applies on top.

## How to check it

- `cargo test --workspace`: `dry_run_tests` (4): an unread object is not one that would be created; same object back is no change, different is a change; a refusal wins over whatever is there; a 502 is not a refusal but a 422 and a 403 are.
- `bun run test`: `YamlEditorDialog.test.tsx` gains four: the server's answer is shown per document; a refused document disables Apply and shows the server's words; a failed dry run falls back to the editor's diff and says so; an unanswered document leaves Apply enabled.
- Live: `tests/live_dry_run.rs` (ignored) runs `dry_run_of` against a real cluster: an edit to a real Deployment is *would change* with `strategy:` and the other defaults present on the server's side and `managedFields` absent; a fresh name is *would be created* and `get` afterwards finds nothing; the narrow identity from #161's harness is never told it would create. The sandbox proxy died before this run completed; the harness is in the PR and I will attach the output when it is back.

## Risk and rollback

Additive on the confirmation only. The apply itself is unchanged. The one new way to be stopped is a document the server refused, which is the server's decision reported earlier rather than later. Reverting removes the command, the panel and the six-way outcome; the editor's diff is what remains, which is what was there before.


belliel:
Live run, which the body said was outstanding. Fresh v1.36 sandbox, test manifests applied, the `narrow` identity from #161's harness header:

```
K8S_GUI_DRY_CONTEXT=killercoda K8S_GUI_REFUSED_CONTEXT=narrow \
  cargo test --test live_dry_run -- --ignored --nocapture --test-threads=1

a_fresh_name_would_be_created_and_nothing_is_stored ... ok
    outcome: Created; get afterwards finds nothing
a_refused_read_is_not_a_creation ... ok
    narrow outcome: Refused { said: "deployments.apps \"expr-demo\" is forbidden: User
      \"system:serviceaccount:k8s-gui-test:narrow\" cannot patch resource \"deployments\" ... Forbidden" }
an_edit_is_a_change_with_the_servers_object_on_both_sides ... ok
    outcome: Configured; `would` carries `replicas: 7`, `live` does not; no managedFields, no status
test result: ok. 3 passed
```

Two things the run taught, both now in the PR:

- The first attempt at the edit case was refused: `spec.selector: Invalid value: ... field is immutable`. The harness had hand-written the Deployment with its own selector, and the server refused it exactly as it would refuse a real apply. That is the feature working on its own test fixture; the harness now starts from the object the cluster holds and changes only the replica count (second commit).
- The sandbox's proxy died mid-run once, and every document came back `Unanswered { said: "502 Bad Gateway ..." }` rather than `Refused`, with Apply left enabled. Before the sixth outcome existed, that 502 would have read as the server saying no.

Dudude-bit:
Ran the review workflow over the branch (4 dimensions, each finding put to independent refutation). What survived is fixed on this branch, and I confirmed both halves with the app open on a kind cluster.

**The preview was one deletion away from being an apply.** `.dry_run()` ran in no test CI executes — all three callers are `#[ignore]`d — so removing it turned the confirmation's preview into a real write with the whole workspace green. On a cluster marked critical that is an apply nobody confirmed. The parameters are their own function now, with a test on the flag.

**A 401 was filed with the 4xx refusals**, so an expired session blocked Apply as though the cluster had rejected the YAML, and put the `CREDENTIALS_EXPIRED:` wire marker on screen as the server's own words. It leaves by the door the rest of the app uses.

**"The current object could not be read" was drawn as a full-green creation diff.** `live` is null both for an object that is absent and for one whose read failed; against an empty string the whole document reads as "this would all be created" — about an object that may be about to be overwritten.

**"Would not change" was decided on the redacted text**, so a rotated `tls.key` compared equal to the old one and a Secret whose whole point changed reported as unchanged, with no diff. The comparison is on the objects; the screen keeps the redaction.

**The promised diff was not drawn.** The sentence for an unanswered document says "this is the editor's own diff", and `would` is null there — an empty promise on the one screen that exists to show a change before it is made. And the preview's key changes only with the buffer, on an observer that never unmounts, so a second apply of the same text was confirmed against the first apply's answer: "would be created" about an object that now exists.

**Where the third state is decided, a test can now reach it.** The 404-versus-refused decision is where `Live::Absent` and `Live::Unread` are actually chosen, and every unit test made both by hand — swapping the arms, which turns "nobody could look" into "there is no such object", changed nothing anybody ran. It is its own function with a test on both arms, and it strips the `READ_DEADLINE` marker, which has no business in prose beside the reader's own language.

PR #173, opened 2026-09-11
feat(logs): freeze an interval so the live tail cannot evict it

belliel:
## What is wrong today

The log pane keeps at most "Keep N" lines and evicts from the head as the stream goes on. Drag a range on the density strip, start reading the burst from four minutes ago, and on a chatty pod the lines under the reader are gone within the minute: the cap does not know which lines are being read. The only ways to hold onto them were to pause the stream, which stops the tail the reader also wants, or to raise Keep to 40 000 and hope. This is the last small item from the plan, "freeze an interval in logs", with its acceptance criterion: live lines do not evict frozen ones, with a test on the budgets.

## What changes

- The buffer takes a frozen interval (`Frozen { from, to }`, in `log-buffer.ts`). Eviction steps over lines inside it, and they are not counted against Keep: the cap bounds what the stream may push out, the freeze is what the reader chose to hold. What can be frozen is what the buffer held at the time, so the whole never exceeds twice the cap. Thawing hands the lines back to the cap, which takes them on the spot rather than on the next batch.
- The time chip gets a ❄ toggle beside its ×. Freezing does not restart the stream and does not touch the lines outside the interval. The freeze outlives the chip on purpose: the chip is about what to show, the freeze about what to keep, so taking the filter off to watch the tail does not throw the held lines away.
- The status bar says `❄ 1 234 frozen lines · thaw` with the range in its title, and is the handle that thaws once the chip is gone. Its meter and "of N kept" now count live lines only, so a buffer holding frozen lines is not drawn as past full.
- The density strip underlines the frozen slices, so the map says what is held.
- Copy in both languages: chip titles and toggle labels say what freezing keeps and what it costs.

## How to check it

- `bun run test src/components/logs/hooks/log-buffer.test.ts`: the frozen lines survive eviction and everything around them goes; live lines never exceed the cap beside them across many batches; a late line inside the interval counts as frozen; thawing evicts down to the cap at once; moving the interval recounts; the field index keeps describing what is retained.
- `bun run test src/components/logs/LogQuery.test.tsx`: the toggle is offered in words, says thaw once frozen, and a chip does not claim a freeze that is not its own.
- In the app: open a pod's logs, drag a range on the strip, press ❄ on the chip, remove the chip, watch the tail run past Keep; the status bar keeps the frozen count and the strip keeps the underline.
- Not rendered here; the components are the existing ones with one control and one segment added.

## Risk and rollback

Memory is bounded by twice Keep, stated in the type's doc. Nothing changes for a reader who never freezes: with no interval the buffer takes the same fast path as before, and the existing buffer tests are untouched. Rollback is a revert.


Dudude-bit:
Reviewed with current `main` merged in (30 commits behind; clean). Five findings; three are fixed and pushed, two are yours to decide because they are design calls rather than slips.

## Fixed

**A freeze could reach into the future.** `commit()` hands `onSelect` the nominal end of the last, still-filling slice — `buckets[hi].start + step - 1` — so the natural drag to the newest slice, "from here to the end", set an upper bound up to one rung ahead, and rungs go to a day. Every line arriving until the wall clock caught up satisfied `isFrozen`, was held outside the cap, and could never be evicted: `dropped` stayed 0 while the buffer grew without limit. Clamped to `Date.now()` — a range that has not happened is not one a reader can choose to keep.

Visible in the chip: dragging to the newest slice now commits `time = 22:00:00–13:56:26`, the wall clock, not the end of the hour.

**Freezing deleted the loaded history.** `room = limit - live.length` still measured the whole buffer against the cap, while `LogStatusBar` two inches away had just been taught that frozen lines sit outside it. Freeze a window with history on screen and every history line vanished — and the history bar then told the reader the *live stream* had filled Keep and they should raise it. One fact, two readers. It is `historyRoom` now, one owner, with the rule under test.

**The freeze outlived the buffer.** A Keep change, Clear, a switch of pod or run all empty it, and the chip went on rendering ❄ with "0 frozen lines · thaw" for an interval whose lines were gone and could never come back. The hook now says when it wipes and the viewer drops the freeze. Checked live: with 4 038 lines frozen, switching Keep 5 000 → 20 000 leaves `16 030 of 20 000 kept` and no frozen claim at all.

## Yours to decide

**Nothing bounds a frozen block even with the clamp.** A long enough interval on a busy pod still outgrows any cap — the clamp stops the *unbounded* case, not the large one. A ceiling in lines, or a refusal to freeze a range wider than some span, is a product decision I did not want to make for you.

**The density strip keeps counting evicted lines.** Once the frozen block pins the head, the bars between the frozen window and the tail go on growing, the header keeps claiming "all N lines landed", `peak` inflates until the live bars collapse to the 2 px minimum, and clicking one of those tall bars jumps to a slice that holds nothing. The strip is the one control CLAUDE.md calls a map that must not quietly stop being a map. The fix depends on what you want the strip to mean once a freeze exists — what the buffer holds, or what arrived.

Dudude-bit:
Second pass, two more fixes pushed.

**Three readers still said the loss is at the head.** Freezing changes what `dropped` means: eviction steps over the frozen lines and takes what is around them, so the missing lines are a hole beside the kept block, not a head the log begins after. The dropped notice, the strip's axis title and its spoken summary all branched on `dropped > 0` and kept saying "older lines have been dropped" — pointing the reader at the left edge, where the frozen block is. There is a `lostLines(dropped, frozen)` returning `"none" | "head" | "aroundKept"` now, computed once in `LogViewer` and passed to both, with its own sentences. Live: the notice reads «Вокруг замороженного интервала отброшено 35 строк. Буфер хранит замороженный интервал и последние 1000 вокруг него».

**The frozen-lines count picked its plural form from a formatted string.** `t("count", "frozenLines", { n: formatCount(frozenLines) })` — `Intl.PluralRules.select` takes a number, `Number("3 472")` is `NaN`, every category falls back to `other`, and Russian then prints the wrong form for every count above 999. That is exactly the magnitude where a formatter starts adding a separator, so it works in every test and fails on a real log. The number itself still printed correctly, which is why nothing looked wrong. Fixed the way the neighbouring `olderLinesDropped` already does it (`n` raw, `{count}` formatted), and `call-sites.test.ts` now fails on any plural entry whose `n` is a formatted value — it found this one and nothing else.

The two design questions from my last comment are unchanged and still yours.

PR #179, opened 2026-09-12
feat(ux): what a week of use asked for: copyable names, a predictable row click, a search that stays, notes after an update

### LINKED ISSUE #178: Мелкие улучшения ;_
Надеюсь еще не надоел своими запросами ;)

1. копирование pod name - около ip аддресов при наведении мышкой появляется иконка копирования. хочется такое для pod name (в заголовке страницы, в таблице?) и возможно для других сущностей.. (есть отдельная кнопка copy name - но кажется она станет лишней). есть копирование по right-mouse-click, но там копируется какая-то внутрення ссылка на http://tauri.localhost/configmaps/ (for example)

2. когда просматриваю список подов, то при клике на один открывается краткая страница с описанием. 
 если кликать на другие поды в таблице, то иногда открывается полная страница. закономерности не нашел ;(
 2.1 можно ли по double-click на строке таблицы открывать полную страницу?
 2.2 если в краткой странице открыта вкладка Logs, то при открытии полной тоже оставаться в Logs ?

3. в Deployments есть краткое описание hpa и pdb, но возможность изменить их я не нашел..

4. Спасибо за работы с файлами. 
 4.1 почему размер ограничен 100мб ?
 4.2 при открытии страницы Files вижу такое
 
Image
но если кликнуть в каталог, то там файл есть и он открывается на просмотр

Image
 ps. этот файл примонтирован из configmap

5. иногда приходится фильтровать объекты по части имени (поды, деплоймент, конфиг-мап), но при переходе между вкладок строка поиска сбрасывается.  можно ли как-то (отдельным чек-боксом?) закреплять?
  можно бы открыть сразу несколько вкладок, например через ctrl+k -> ctrl+enter, но там тоже сразу после    открытия вкладки закрывается поиск.

6. на странице подов есть колонка namespace - зачем?   сверху и так видно. а если выбрать несколько namespace, то колонки нету...

pss. это просто замечания от использования, если что-то было так и задумано, то может просто я не в курсе
psss. кажется не хватает доки с фичами или whats new при обновлении. а то не все ходят в github..

belliel:
Closes #178.

## What is wrong today

A reader who used the app for a week wrote down what got in the way. Each item is small; together they are the difference between a tool that is used and one that is put up with.

- A pod's name could not be copied where it is read: the hover mark exists for addresses only, and "copy link address" in the right-click menu copied `http://tauri.localhost/...`, which opens nothing anywhere.
- Clicking a pod in a list sometimes opened the peek and sometimes the page, with no rule the reader could find. There was one (the name link peeks, the whitespace navigates), and it was invisible. Leaving a peek's Logs tab for the page landed on Overview.
- The Deployment page described its autoscaler and its budget and gave no way to change either; neither kind has a page.
- The Files tab opened on a ConfigMap key mounted over a single file and said the path could not be opened. Downloads over 100 MiB were refused with "refused in this version" and nothing else.
- The search box emptied whenever the reader left the tab and came back, and the palette closed after opening a hit in a background tab, so opening three pods meant typing the search three times.
- With one namespace chosen the pods list still carried a namespace column that said the same word on every row.
- Nothing in the app said what a new version brought.

## What changes

- **Names.** The same hover mark addresses have, on the page title and in every list row. A right-click on any object name opens the app's own menu: copy name, copy link (a `rubick://` link that opens the same place), open in a new tab.
- **Rows.** A plain click anywhere on a row peeks, the same as the name; a double click or Enter opens the page; modified clicks open tabs as before. Only rows whose route is an object behave this way, so a table of Helm releases is a plain link still. Leaving a peek for the page carries the open tab with it.
- **Autoscalers and budgets.** Each one named on a workload page carries the YAML editor every detail page has, through the manifest command that already serves any registered kind.
- **Files.** A mount with a `subPath` opens the tab on its folder, where the file is a row with its mount tag. The download cap moves to 2 GiB, and anything over 100 MiB is asked about first, naming the size and the wait; the idle timeout still ends a stream that goes quiet.
- **Search and the palette.** The list search lives in the query string, which is what a tab records, so it survives leaving and returning. The palette stays open after a background tab. A ctrl-click on the name inside a palette hit also opened the object twice, once for the link and once for the row; it opens once now.
- **The namespace column** is hidden while exactly one namespace is chosen. Several are grouped under captions already; all namespaces keep the column.
- **Release notes.** The first launch of a new version opens what arrived since the version last seen, parsed from `CHANGELOG.md` at build time; a first install records where it starts and opens nothing. Settings › About has a "What's new" row that opens the current version's notes.

## How to check it

- `bun run test`: the new cases name the issue item they cover. Files: the start path for a single-file mount, the confirmation before a big download and its absence for a small one, both halves of `shared/file-limits.json`. Rows: peek on click, page on double click, plain link where the route is not an object, the peek's tab travelling. Names: the right-click reaches the app's menu at the pointer, the copy mark keeps the row and a double click out of it. Search: read from and written to the query string; the palette stays open after a background tab; the namespace column hidden for one namespace and back for none. Governance: the editor beside the autoscaler. Release notes: opened after an update with every skipped release, not on a first launch, closed and recorded.
- `cargo test --workspace files::`: the Rust side of the shared limits.
- In the app: right-click a pod name; click a row, then double-click one; open a peek's Logs tab and press Enter; search a list, switch tabs and come back; pick one namespace; open a Deployment with an HPA; open Settings › About › What's new.

## Risk and rollback

The row click is the one behaviour change a reader will feel on day one: the list peeks where it used to navigate. Back closes the peek, and the page is a double click away. Everything else is additive. Rollback is a revert.

### DIFF
--- a/shared/file-limits.json
+++ b/shared/file-limits.json
@@ -1,5 +1,6 @@
 {
-  "downloadMaxBytes": 104857600,
+  "downloadMaxBytes": 2147483648,
+  "downloadConfirmBytes": 104857600,
   "previewMaxBytes": 524288,
   "maxEntries": 20000
 }
--- a/src-tauri/src/files/mod.rs
+++ b/src-tauri/src/files/mod.rs
@@ -29,8 +29,9 @@ pub use parse::{FileEntry, FileKind};
 pub const PREVIEW_MAX_BYTES: usize = 512 * 1024;
 /// The cap a download is cut off at — counted as the bytes arrive, into a
 /// scratch file beside the destination, so nothing over it is ever renamed
-/// onto the reader's own file.
-pub const DOWNLOAD_MAX_BYTES: u64 = 100 * 1024 * 1024;
+/// onto the reader's own file. Wide, because the exec channel is the only
+/// slow part and the frontend asks before starting anything that large.
+pub const DOWNLOAD_MAX_BYTES: u64 = 2 * 1024 * 1024 * 1024;
 /// Rows per event on the way to the frontend.
 pub const BATCH_ROWS: usize = 500;
 /// The most rows one listing hands over. Nothing capped this: `list_dir`
--- a/src/components/files/FilesTab.test.tsx
+++ b/src/components/files/FilesTab.test.tsx
@@ -318,6 +318,60 @@ describe("FilesTab", () => {
   });
 
   /**
+   * Issue #178: a file over the old 100 MiB cap was refused outright. It is
+   * asked about now, and the file dialog opens only after the answer. Would
+   * break if the confirmation went away, or if the cap fell back under it.
+   */
+  it("asks before a download that will take minutes, and not before a small one", async () => {
+    listing.mockReturnValue(
+      done([
+        {
+          name: "core.1842",
+          kind: "file",
+          mode: "600",
+          size: 300 * 1024 * 1024,
+          modified: null,
+          owner: "app",
+          group: "app",
+          target: null,
+        },
+      ])
+    );
+    readContainerFile.mockResolvedValue({
+      state: "preview",
+      preview: {
+        bytesRead: 1024,
+        truncated: true,
+        binary: true,
+        nonTextShare: 0.5,
+        lossy: false,
+        text: null,
+      },
+    });
+    const { save } = await import("@tauri-apps/plugin-dialog");
+    wrap(
+      <FilesTab
+        pod={pod()}
+        via={null}
+        onDebug={() => {}}
+        onStopVia={() => {}}
+      />
+    );
+    await userEvent.click(screen.getByText("core.1842"));
+    await userEvent.click(
+      await screen.findByRole("button", { name: "Download" })
+    );
+    expect(
+      await screen.findByText(/Download core.1842 \(300.0 MB\)\?/)
+    ).toBeInTheDocument();
+    expect(save).not.toHaveBeenCalled();
+    await userEvent.click(
+      screen.getAllByRole("button", { name: "Download" }).at(-1)!
+    );
+    await waitFor(() => expect(save).toHaveBeenCalledTimes(1));
+  });
+
+  /**
    * "Reading, and nothing has arrived yet" and "the tool finished and found
    * nothing" are two different answers. Only the second one is emptiness, and
    * a listing that streams its rows in spends every read in the first.
--- a/src/components/files/FilesTab.tsx
+++ b/src/components/files/FilesTab.tsx
@@ -30,6 +30,7 @@ import {
   matches,
   modeText,
   mountFor,
+  startPath,
   parentOf,
   sortEntries,
   type FileEntry,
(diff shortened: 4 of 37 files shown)


Dudude-bit:
Reviewed against the six items of #178, in the app on a live cluster and with
an adversarial pass (7 reviewers by dimension, 3 refuters per finding, majority
to kill: 59 findings, 19 survived). Pushed the merge plus the fixes.

## Every item of #178 is closed

Checked each in the running app, not by reading:

1. **Copy a name.** Hover mark on the name in the table, the app's own
   right-click menu, and `rubick://open/kind-rubick-gui/pods/...` in the
   clipboard instead of `http://tauri.localhost/...`.
2. **A click means one thing.** Always the peek, wherever on the row.
   **2.2** — opened the peek's Logs, pressed Open full page, landed on Logs.
3. **HPA and PDB.** "Edit YAML" beside `hpa-busy`; the real manifest opens with
   Validate and Apply.
4. **4.2** — built a pod with a `subPath` mount (there was no fixture for the
   shape): the tab opens on `/etc/app` and the file is a row tagged
   `from subpath-demo`. **4.1** — 2 GiB and the 100 MiB confirm are in
   `shared/file-limits.json`.
5. Search survived leaving the tab and returning; the palette stayed open with
   its query after `⌘↵`.
6. Namespace column gone with one namespace chosen.

The release-notes dialog reads well, which answers the postscript.

## Eight defects, fixed here

**The right-click menu opened nowhere near the pointer** — a sidebar's width
right and a header's height down, drifting further with the scroll. The anchor
is `position: fixed` inside the page container, which carries a transform, so
`fixed` measured from that box and not the window. Anchored on the body now.
This is the one I would have shipped and heard about within a day.

**The menu never reached a custom resource, a Helm release or a CRD.** The fix
went onto `ObjectLink`; those three render `RouteLink`, which had no handler —
so the largest family of rows in the app still opened the webview's menu with
the `tauri.localhost` link. That is item 1, unclosed where it matters most.

**A double click on a row's own name did nothing.** `openPage` bailed on
`closest("a")`, and the name is an anchor — so the whitespace opened the page
and the name, where the eye goes, only peeked. Now a link pointing at the row's
own href navigates; a link pointing elsewhere (a node, an owner) keeps its
meaning.

**The CRDs list answered a click two ways inside one row** — the row peeked
(its href resolves to a peek target) and the Kind cell navigated. That is item
2's complaint, recreated one page over. The cell is an `ObjectLink` now.

**`?tab=children` is a tab no page has.** The peek has one `children` tab for a
workload's pods and a CronJob's jobs; each page names it after the kind. So
leaving a Deployment peek's Pods tab landed on Overview *and* left a parameter
the scope tab recorded as the route, replaying the wrong landing. There is a
`pageTab` translation now. Related: **a custom resource's page never read
`?tab=` at all** — it held its own `useState("overview")`.

**The search box was seeded from the query string once.** Click the sidebar row
for the list you are already on and the address loses `?q=`, while the box and
the rows keep the old filter — then the filter is silently gone on the way
back, which is the failure item 5 reported. The query string is the authority
now and the box follows it.

**A double click in the quick-actions gutter navigated**, where a single click
there is deliberately inert.

## Two more the app showed

**The release notes printed literal backticks.** `**\`AzureIdentity\` was
counted as…**` renders with the backticks visible, because the inline renderer
is flat. Line 36 of the CHANGELOG this parses is exactly that shape, so it is
in the first release the feature will ever show. Bold and emphasis recurse now.

**The right-click test passed with its `preventDefault` deleted** — it asserted
the store was filled and not that the event was claimed, and claiming it is the
whole of what stops the webview menu. It asserts both now.

## Smaller

`ObjectMenuTarget` extended `PeekTarget` and carried a kind, a namespace and a
crd the menu never read — three things every caller had to supply, which is
part of why the menu never reached the links that have no kind to give. It
carries a name and an address now. "Copy link" also confirmed with
`Copied where you are`, the sentence written for copying the current view.

## The merge

The branch was 24 commits behind and conflicted: main deleted `searchKey` in
the search refactor (#188) while this PR added `searchParam` beside it. Merged,
kept `searchParam`.

Every fix has a test, and each test was sabotaged to confirm it fails for its
own reason. Gates green: 2757 frontend, 638 Rust, clippy at zero.

PR #176, opened 2026-09-11
feat(prometheus): one row for the address and the operator's objects, and monitors read as a story

belliel:
## What is wrong today

A ServiceMonitor is configuration for a scrape that happens somewhere else, and the app showed it as a generic custom resource: a name, a spec, nothing about whether it works. The three questions a reader has about one could not be answered from the object: does its selector select anything, does any Prometheus pick it up, and is the connected Prometheus actually scraping what it selects.

Two more things were wrong once the first version of this branch existed. Integrations listed "Prometheus" and "Prometheus Operator" as two rows, and nobody could say from the sidebar why there were two. And on a cluster with a partial install (the Prometheus CRD missing, or the list of Prometheus objects refused) the page collapsed the unread list into an empty one and painted every monitor red with "no Prometheus picks it up", which is the defect class this repository exists to stop.

## What changes

- **One vendor.** `prometheus` is detected by any of the operator's three CRDs and connected by an address, and is present by either. The page has two tabs: **Monitors** (the operator's objects, shown when they exist) and **Connection** (what the address answers, always, since it is where "not connected" is said). The sidebar row carries the monitor count with the worst tone as its dot. The registry rule "configured means never detected" is gone: `pageDecision` opens the page on an address even when the scan found no kinds, and answers "not connected" rather than "not installed" for a vendor whose other way in is an address. `Loki`, the only other configured vendor, is unchanged in behaviour.
- **Every operator kind reads as read, unread or absent**, from the CRD list. A refused Prometheus list makes pick-up unknown on every row (a warning, with the reason), a missing Prometheus CRD makes pick-up not judged and is said once in the header, and a monitor with no pool that nothing is known to pick up is waiting rather than fine. A missing PodMonitor CRD is a grey line, not a warning. The ServiceMonitor list is read the same way, so the page lives without it.
[...]

PR #182, opened 2026-09-13
feat(prometheus): an Alerts tab that says which rules are picked up, loaded and firing, and what is firing about an object

belliel:
Stacked on #176 (the merged Prometheus vendor); the diff shrinks to this branch's one commit once that lands.

## What is wrong today

The operator has four kinds and the app read three. `PrometheusRule` is where alerting lives, and it was a generic custom resource here: a name, a spec, nothing about whether any Prometheus picks it up, whether the rules loaded, or what is firing. The three questions a reader has about an alert (is it evaluated at all, is it firing, what about) were answered only in the Prometheus UI. And on a pod page nothing said that an alert had been firing about that pod for the last hour.

## What changes

- **A new backend read**, `prometheus_rules`, on the existing tier-3 connection: `/api/v1/rules?type=alert`, parsed into groups, files, state, health, Prometheus's `lastError` verbatim, and each rule's active alerts with their labels. A refused list is an error, never no rules.
- **An Alerts tab** on the Prometheus page, present when the `PrometheusRule` CRD is, in the same shape as Monitors. Each object is answered three ways, each with its third state:
  1. **Picked up by**: `ruleSelector` and `ruleNamespaceSelector`, with the monitors' absent-versus-empty reading, shared through one `selectedBy` rather than a second copy of the loop.
  2. **Loaded**: the connected Prometheus has a rule file named after the object (`<ns>-<name>.yaml`, or `<ns>-<name>-<uid>.yaml` on the releases that append the uid). Not loaded while picked up is a finding; not loaded while nothing picks it up is the same fact said once. A rule the object has that Prometheus does not is named.
  3. **Rules**: every alerting rule with its state, `for`, severity, summary, expression (copy, open in Prometheus), an evaluation error in Prometheus's words, and the active alerts as rows that link to the object each one names.
- **A ladder** grouped firing, broken, pending, quiet, or not checked without a connection; firing outranks a fault, since the alert is what the reader came for and the fault is why the rest is silent. Selection lives in `?rule=`.
- **A new capability**, `alerts.about`, provided by the Prometheus vendor and consumed by every pod and workload page through `useCapabilityState`: the alerts firing or pending about the object, matched by the label kube-state-metrics writes for its kind and, for a workload, by the pods it runs; the matching label is shown, so a guess at a pod's owner reads as one. Absent capability draws nothing: an empty block would say "no alerts", and no evaluator is not that.
- Copy in both languages under `alerts`.

## How to check it

- `bun run test src/integrations/prometheus`: reading rule objects (alerting rules kept with their group, recording rules counted apart), pick-up by the rule selectors, ownership of a file in both spellings, pairing rules with what Prometheus loaded, the order and content of findings (firing first, then faults, then pending), not-loaded only when something is known to pick the object up, and `alertsAbout` by kind label, by pod name pattern, by namespace, and for a node.
- `cargo test --workspace prometheus::`: the rules parser keeps the file, the state, the alerts and Prometheus's words, drops recording rules, and refuses a refused list.
- Rendered with the real frontend on a fake backend fed by the recorded kube-prometheus-stack fixture plus synthesized rule objects and `/api/v1/rules` answers: a firing object with pod links and a pending alert, an object whose rule fails to evaluate, one nothing picks up, one not loaded; dark and light.
- **Not yet checked live:** the rule file naming against a running operator. Both spellings are accepted; a sandbox with kube-prometheus-stack settles which one v0.85 writes.

## Risk and rollback

A cluster without the `PrometheusRule` CRD sees no tab and no block. With one and no connection, the tab reads pick-up from the objects and says the rest is unchecked. The pod and workload pages make one extra read of `/api/v1/rules` per list refresh while a Prometheus is connected. Rollback is a revert.

## Summary by CodeRabbit

* **Новые возможности**
  * Добавлена страница Alerts для просмотра правил Prometheus, их состояний, ошибок вычисления и активных срабатываний.
  * Реализованы фильтрация, группировка, поиск связанных Kubernetes-объектов и переходы к Prometheus и исходным ресурсам.
  * В карточках Pod, Deployment, StatefulSet, DaemonSet, ReplicaSet, Job и CronJob отображаются связанные предупреждения.
  * Добавлена вкладка Alerts со счётчиками правил и активных срабатываний.
  * Добавлена локализация интерфейса на русском и английском языках.

* **Тесты**
  * Добавлены проверки обработки правил, состояний, ошибок подключения и связывания с Kubernetes-объектами.


belliel:
Checked live against kube-prometheus-stack on a v1.36.1 sandbox (operator v0.85.0, Prometheus v3.14.0, `helm install --skip-crds` after server-side apply of the CRDs), 2026-09-13 01:02–01:12Z.

- **The rule file is `<namespace>-<name>-<uid>.yaml`** on v0.85: `k8s-gui-test-shop-alerts-f265ea35-….yaml`. Both spellings were accepted; the uid one is the real one, and the recorded test now proves it against every picked-up chart object.
- **Three objects written for the page behaved as designed:** `shop-alerts` (label `release: kps`) is picked up, loaded, and `ShopAlwaysFiring` fires; `nobody-picks-up-rules` (no chart label) is selected by no Prometheus and reads as not picked up, once; `broken-rules` (duplicate labelset) comes back `health: err` with Prometheus's words, `vector contains metrics with the same labelset after applying alert labels`.
- **One thing the recording taught, fixed in c5d2ea9:** `/api/v1/rules?type=alert` lists no group made of recording rules only, so an object like `k8s.rules.container-cpu-usage-seconds` has no file there and nothing wrong with it. It now reads as recording-only rather than not loaded.
- **The object side:** a pod written to crash (`crash-demo`) is named by `ShopPodCrashLooping` (firing, critical) and by the chart's `KubePodCrashLooping` and `KubePodNotReady` (pending), all through the `pod` label, and not by anything in another namespace. The Watchdog names no object.
- Over the 38 rule objects the chart writes: 3 firing, 2 broken (the two written to be), 5 pending (the kubeadm control-plane monitors, the same ones the Monitors tab shows down), 28 quiet.

The recording is in `src/lib/__fixtures__/live-prometheus-alerts.json` and `alerts/model.live.test.ts` reads it the way the page does.

Dudude-bit:
Rebuilt on current main — the branch carried the monitors work that landed as #176 plus the fixes made to it since, so only its own five commits are replayed — and reviewed.

**Three verdicts stated from what the tab had not read.** Prometheus writes three healths and the app read one: a rule it has loaded and never evaluated answers `unknown`, which fell through to the green "loaded and evaluating" — about a rule that has never run. `pickedUpUnknown` was in no group at all, so a row whose pick-up nobody could judge was filed under Quiet, in green, with the word "unknown" beside a check mark. And "picked up, but the connected Prometheus has not loaded it" was asserted while the pick-up was unknown, which is a claim about a Prometheus nobody identified.

**The workload page drew byte-identical nothing** for "nothing is firing" and "the read failed" — on the page where "no alert" is the fact a reader leans on. `AlertsAbout` now says which it is, and has its first test file.

**The tab mark and the sidebar dot** were built from the firing alerts alone, so a rule object nothing picks up, or one Prometheus cannot load, left a plain count on the tab and nothing at all on the rail beside a page that is all red. Both ask the same rows the page draws.

**Two numbers from two populations:** "{n} alerts firing across {rules} rule objects" counted alerts over everything the connected Prometheus has loaded and rules over the listed PrometheusRules.

**And the parser's own test** asserted neither of the two fields both readers key on — blanking an alert's `state` or a rule's `name` left the whole Rust suite green. The name is how a spec is matched to what Prometheus loaded; the state is the difference between firing and pending on the page.

Smaller: an alert whose subject is a HorizontalPodAutoscaler was drawn as a link to a page this app does not have — the kind is in the label table and not in the routable one.

Gates: tsc, lint, 3087 frontend tests, 723 Rust tests, clippy, both formatters, comment budget 3%.

Dudude-bit:
A second pass over what the review left, and a live check against kube-prometheus-stack on kind.

**One read per cluster, not one per object.** `alerts.about` handed back the whole rules collection for every object a reader opened — every page, every peek, every poll. It is a read and a pick now: the surface reads once per cluster and asks for its slice.

**The peek carries it too.** The detail page showed what was firing about a workload while the peek at the same workload showed nothing — and the peek is where somebody looks first. A kind no evaluator labels gets neither the block nor a "could not be read" sentence over it.

**The link is the supplier's.** "All alerts" was a spelled-out `/integrations/prometheus?tab=alerts` from outside the seam; the vendor now says where its alerts are answered and the block reads that, so it points at whoever answers. And the Russian line describing what this integration gives still mentioned only monitors.

PR #183, opened 2026-09-13
feat(app): every key the app answers to, on one sheet behind ?

belliel:
## What is wrong today

The status bar hints at three shortcuts and nothing says the rest exist. Copy a link is `mod+shift+C`, tabs answer to `mod+T`, `mod+W`, `ctrl+Tab` and `mod+1…9`, the log viewer solos a container on a digit, the density strip selects a range with shift and the arrows. A reader learns these from a colleague or not at all.

## What changes

- `?` anywhere outside a text field, a terminal or an editable region opens one sheet with every key the app answers to, grouped by where it works: everywhere, tabs, search, lists, the peek panel, the Files tab, logs. `?` again or Escape closes it. The status bar gains a fourth hint, "? keys", that opens the same sheet by click.
- The table behind it, `src/lib/shortcuts.ts`, is the one place a shortcut is written down a second time, and every row names a handler that exists; nothing was invented for the sheet. Keys are logical (`mod+K`) and rendered per platform by the keycap component the app already had.

## How to check it

- `bun run test src/components/layout/ShortcutsOverlay.test.tsx`: `?` opens the sheet outside a text field and not inside one, `?` again closes it, and every listed key renders on both platforms without a raw `mod` leaking through.
- In the app: press `?` on any page; press it inside the search box and see a question mark typed instead.
- Rendered on the fake backend, dark and light.

## Risk and rollback

Additive: one window listener and one dialog. `?` was bound to nothing before. Rollback is a revert.


belliel:
Duplicate of #166, which already covers the same sheet with a completeness test over the handlers. Closing this one.

PR #184, opened 2026-09-13
feat(app): why slow, said from what the app measures for free

belliel:
## What is wrong today

On a big cluster the app stutters and says nothing about it. The performance recorder in Settings › Diagnostics knows exactly what happened, and it is off by default for a good reason: while it runs every backend answer is serialised a second time to count its bytes. So the reader who feels the stall is the one with no numbers, and "why is this slow" is answered by guessing.

## What changes

- **The cheap half of the recorder is always on.** A small stall watch takes the same long-task observer the recorder uses (or late frames where the webview has none), the row count of every backend answer (an array's length, never a walk over the value), and the lists on screen (a table reports its size and takes it back when it leaves). Nothing under a thousand rows and nothing older than a minute is kept.
- **A count in the status bar** while there is something to say: "3 stalls", in the warning tone, red past 250 ms or five stalls. Absent otherwise: a row reading "no stalls" all day would be the app praising itself.
- **Opening it names the facts:** how many stalls and the longest, with which observer said so; the big lists on screen and why they cost (every watch batch is filtered and sorted over the whole list, a search re-reads every row); the largest answer and its command. Then what to narrow, and a button into the recorder for timings per command and per render.
- `startFrameWatch` now takes any task sink, so the recorder and the watch share one observer implementation instead of two.

## How to check it

- `bun run test src/lib/stall-watch.test.ts src/components/layout/StallIndicator.test.tsx`: the window slides and the longest stall is named; only big answers are remembered, by length; big tables are listed largest first and forgotten on unmount; the indicator is absent without a stall, appears with one, and the sheet names the list, the command and the stall.
- In the app on the perf rig (`make perf-rig`): open the pods list, watch the status bar; open the sheet.
- Rendered on the fake backend with a 1500-row list and a forced 320 ms stall, dark and light.

## Risk and rollback

One `requestAnimationFrame` loop where the webview has no long-task observer, one `Array.isArray` per command answer, one effect per mounted table. Rollback is a revert.


Dudude-bit:
Reviewed with current `main` merged in (17 commits behind; one conflict in `data-table.tsx`, both import blocks belonged, resolved).

**A hidden or minimised window came back as a multi-second stall.** On a webview without the long-task observer the watch falls back to the rAF loop, and rAF does not run while the page is hidden, minimised or occluded — so the first frame back carries the whole absence. Minimise for five minutes and the status bar shows a red turtle and the sheet reads "1 stall, the longest 299 983 ms", then advises narrowing the namespace scope. A five-minute freeze that never happened.

Before this PR the loop ran only inside a recording somebody had started and was watching; `App.tsx` now runs it for the life of the process, so the artefact would have been on every install, unasked.

The gap across a return is dropped rather than guessed at — a period the watch could not look at the main thread is not a measurement of it — and only that one gap: a window that had once been minimised must not quietly stop reporting stalls. Both halves are under test and both sabotages fail the right one.

Dudude-bit:
Second pass, three more fixes pushed.

**The report answered for channels it never looked at.** Both rows read "None over a thousand rows." when empty — a confident negative about a question they cannot answer. `wrapCommand` and `DataTable` feed them and nothing else, so log lines and watch batches can block the thread without reaching either. Each row now names its own scope, and when both come back empty a line underneath says what is not counted. Checked live on the "1 stall" turtle.

**An English noun reached the Russian sentence.** `rowLabel ?? "rows"` hard-coded the word for a table that wears no label, so a Russian reader got "1 234 строк rows". The label is now `string | null` end to end, and a label-less table gets a sentence with no noun in it. A kind's plural still goes out untranslated, which is correct.

**The threshold had no test that could fail on it.** The counts in the existing cases sit far from 1 000, so moving `BIG_LIST_ROWS` either way left them green. There is a boundary case written against the constants now — at the threshold counts, one row under does not — plus one for a list that falls back under it and stops being named.

The always-on cost on the frame-gap fallback is still worth your judgement; I have not touched it.

PR #202, opened 2026-09-15
NetworkPolicies, read as what they do

Dudude-bit:
Rubick did not read core `NetworkPolicy` at all — no list, no detail page, no
peek. This adds all three, plus the reading that makes them worth having.

A NetworkPolicy is the one core kind where the shape of the object and its
effect come apart, and every pair below is a pair a manifest will not tell
you apart:

- `ingress: []` on a governed direction **denies everything**. `ingress: [{}]`
  is a rule with no peers under it and **allows everything**. One character of
  YAML, opposite meanings. The row says which.
- A direction that `policyTypes` does not name is one this policy **makes no
  claim about**, and another policy may govern it. Drawn as "says nothing",
  never as denied.
- `podSelector: {}` is **every pod in the namespace** — the widest thing a
  policy can say, and as a blank cell it would read as the narrowest row on
  the page.
- Two selectors inside **one** peer are an AND (those pods *in* those
  namespaces); as two peers they are an OR and far wider. Drawn as one line
  and two lines respectively.
- An absent `namespaceSelector` is the policy's own namespace — the narrowest
  peer there is — and an empty one is every namespace in the cluster, the
  widest. The same three shapes mean opposite things on the two axes, which is
  why each axis is read by its own function.

## The count, and the read that did not happen

The column this page exists for is **how many pods each policy actually
picks**. A policy whose selector matches nothing is accepted by the API
server, listed like any other, and enforces nothing; no other screen in this
app can say so, because the selector is in one object and the labels are in
another.

Which makes the refusal the important case. A reader with `list
networkpolicies` and without `list pods` is an ordinary RBAC shape, and for
them the count is `None`, never `0` — `0` is the finding, and handing it to
somebody who merely cannot see pods would invent it. Verified on a live
cluster with exactly that role: every row read "pods not read", faint, and
not one said "no pods" — including the policy that really does select none.

## Tests

Nine Rust cases against specs recorded from a real 1.36 API server, and nine
frontend cases over the model. Each was sabotaged and confirmed to fail for
its own reason.

`test-manifests/k8s-gui-all.yaml` gains six policies, one per reading above,
so the next person has the specimens.

## What else had to move

`peek-sources.tsx` needed an entry: without one the panel fell through to the
generic manifest walker, which flattens `ingress.0.from.0.namespaceSelector`
into a dotted path and loses the AND. The verdict words, the port text and the
peer sentence are shared by the list, the page and the peek, so the three
cannot drift.


Dudude-bit:
Ran an adversarial review over the branch: seven reviewers by dimension, then
three refuters per finding with a majority needed to kill it. 32 findings, 20
survived, which dedupe to eight real defects. All eight are fixed in 94f4ff0
and each was checked on screen against a live cluster.

**"allows all" for a direction that only opens named ports.**
`opens_to_everything` read peers and ignored ports, so `egress: [{ports:
[5432]}]` — reach anything, but only on 5432 — was summarised as wide open in
warn amber on the column you scan the list with. It now needs a rule that
names neither peers nor ports.

**A port entry naming only a protocol printed `null`.** `port` is optional and
the API server does not fill it in; `{protocol: UDP}` means *every* UDP port,
and `portText` rendered "UDP/null". Now "every UDP port".

**The peek drew rules for a direction the policy does not govern.** The object
keeps an `ingress:` block that `policyTypes` does not name, and the panel
listed its peers — telling a reader ingress was restricted to them while the
list row said "says nothing" and the detail page hid the direction. Ingress
was in fact wide open. This is the page ↔ peek pair CLAUDE.md names, with the
peek giving the more dangerous answer.

**`isIntersection` was simply wrong.** The two selectors of one peer are
*always* an AND; the OR is between peers, and peers are already separate rows.
So `{namespaceSelector: {tier: trusted}}` rendered through the union string as
"every pod · tier=trusted", which reads as unscoped — the direction that opens
the cluster. The function and the union string are gone; every peer now reads
"{pods} in {namespaces}".

**Russian said «откуда угодно» on egress**, inverting the direction: the
English "anywhere" is neutral both ways and the Russian is not. Split into
`fromAnywhere` / `toAnywhere`.

**«в все пространства имён»** — the wrong case, on the widest peer a policy can
name. The namespace half now carries the preposition's case
(«в любом неймспейсе»).

Two smaller ones: the Rules tab counted rules it then refused to draw, and the
header badge dropped the warn tone that `directionFact` returns and that the
row and the Overview fact below it both painted.

Four new tests, each sabotaged: the port-scoped verdict, the port-less entry,
and a new `peek-network-policy.test.tsx` covering the ungoverned direction and
the three pod-count answers — the peek source had no test at all, which is how
it drifted. One new fixture, `ports-without-peers`, carries both port shapes.

Dudude-bit:
Second pass over the twelve findings the refuters voted down, because refuters
err in both directions. Four of them were real, fixed in 08072f9:

**`NetworkPolicy` was missing from the delivery `API_GROUPS` table.** So
`apiGroupOf` answered `null`, `deliveryScopeOf` returned no column, and every
GitOps surface on the new pages — the list column, the detail block, the peek —
was silently empty, while `Ingress` in the same API group works. Nothing
cross-checked that hand-kept table against the registry, which already states
each kind's `apiVersion`. There is a guard now.

It immediately found nine more kinds already missing: `Gateway`,
`GatewayClass`, the five route kinds, `HorizontalPodAutoscaler` and
`PodDisruptionBudget`. Those are pre-existing and not this PR's scope, so they
are listed by name in the guard as a `NOT_YET` set rather than quietly
excused — a guard that skipped them would read as "every kind is covered",
which is the shape of the bug it guards against. Worth a follow-up.

**The peek panel had no Delete**, though `delete_network_policy` shipped with
the diff and the panel offers one for every other kind that has a delete.

**The detail page never passed `createdAt`**, so the age the list row and the
peek both show vanished on the object's own page. The list ↔ detail pair.

**Both commands pulled whole pod bodies on every poll** to read labels.
`list_metadata` is what `overview.rs` and `helm/secret.rs` already use for
exactly this; checked live that the counts are unchanged (2 / 29 / 1 / 2 / 2 /
no pods / 2, identical before and after).

The other eight refutations hold. The performance ones about IPC size were
right that pods never cross the boundary — only the count does — which is why
the fix above is about the API-server fetch rather than the message.

PR #206, opened 2026-09-15
Three tables keyed by kind that answer with silence

Dudude-bit:
The delivery table held sixteen of the registry's twenty-nine kinds. A kind
missing from it makes `apiGroupOf` answer `null`, and every caller reads that
as "no delivery to speak of" — the list column, the detail block and the peek
marks go quiet together, with nothing failing anywhere.

The seven Gateway API kinds sat like that on three detail pages already
written to ask: `GatewayDetail`, `GatewayClassDetail` and `GatewayRouteDetail`
each call `deliveryOfKind` and each got `null` back, purely because the group
was never written down.

**A kind's group is a fact about the API, not an opinion about this app**, so
all thirteen missing kinds are in and the guard has no exception list. Whether
a *list* draws the column stays the separate, editorial question it was, and
`MADE_BY_THE_CLUSTER` gains Event and Node — nobody commits either to a
repository. `Namespace` deliberately does not join them: namespaces are among
the first things a GitOps repo holds, and the one nobody delivered is the
interesting row.

## Checked, not reasoned about

On a kind cluster with the Gateway API CRDs installed, a Flux `Kustomization`
named `apps`, and `gwtest-edge` carrying the Flux labels:

- the **Gateways list** grew a Delivery column: `gwtest-edge` reads "labelled,
  not listed" in amber, the other two "not delivered"
- the **Gateway detail page** shows the full banner — *"Labelled as delivered
  by apps, which does not list it. The label is a claim anybody can write, and
  the Kustomization it names does not have this object in its inventory."*
- the **HTTPRoute** detail page and peek both show `Flux · apps · unconfirmed`
- a **Namespace** carries the same marks in the peek, which is how that entry
  earns its place with no list asking for a column

Every one of those said nothing at all before.

## The guard

`delivery.test.ts` now holds the table to the registry with no `NOT_YET` set —
any kind it would have to excuse is one the table should simply name. Both
halves are sabotage-checked: dropping a Gateway kind, writing `autoscaling/v2`
where the group belongs, turning the column back on for Event and Node, and
excluding Namespace each fail the test they should.

## Not in scope

`GatewayRoutesList` is a bespoke board for five kinds, not a `ResourceList`,
and `ResourceList.delivery` takes one `{group, kind}` for the whole table. A
Delivery column there needs a per-row kind and is its own change.


Dudude-bit:
Pushed a second commit: the same shape of gap, one table over.

`deleteGateway`, `deleteGatewayClass` and `deleteGatewayRoute` have existed
for releases, and not one of the seven Gateway API kinds was in the peek's
`DELETE_COMMANDS` — so the panel offered no Delete while the detail page
deleted them perfectly well. No error, nothing to notice. Exactly how
`NetworkPolicy` shipped last week.

The guard names the `delete_*` each kind would use and asserts the generated
bindings export it, so a command landing without a peek entry fails; kinds
that deliberately have none (Node is drained, a Namespace takes everything
with it, Event and ReplicaSet are the cluster's) are listed rather than left
looking like an oversight. Both halves sabotage-checked.

Three tables keyed by kind now fail by going quiet rather than breaking —
delivery groups, `ROUTABLE`, and this one — so CLAUDE.md's "adding a resource
kind" checklist names all three and says what each one's silence looks like.
That checklist is what would have caught both of these before review did.

PR #214, opened 2026-09-19
fix(tell): a restart is not finished the moment it is requested

Dudude-bit:
Findings from an adversarial review of #211 and #212, which I merged without running one. Four survived refutation and are fixed here; the review is also how I learned it had independently found both of the Codex comments, ranked just below my verification cap.

### A restarted StatefulSet or DaemonSet was told "rolled out" before a pod moved

The worst of it, and mine. `settled` for these two kinds was `ready === desired && current === desired` — counts that are *already* satisfied the instant the template changes, because the controller has only requested the first deletion. The watch I added in #212 passes `generationBefore`, so `acknowledged()` is true on the first status write, and the verdict rests entirely on `settled && caughtUp`. A Deployment escapes because its arm also asks `replicas.updated === desired`; `StatefulSetInfo` and `DaemonSetInfo` carried no such count at all, so nothing in the payload could tell "has not started" from "finished".

Under `updateStrategy: OnDelete` it is not a race but a certainty. Verified on kind: `kubectl rollout restart` on an OnDelete StatefulSet bumps the generation, `observedGeneration` catches up, `ready` and `current` sit at `desired` — and the pod is never replaced. (The review claimed kubectl *refuses* this; it does not, on v1.36. It restarts happily and rolls nothing, which is worse.)

`StatefulSetReplicaInfo` now carries `updated` (`status.updatedReplicas`) and `DaemonSetInfo` carries `updated` (`status.updatedNumberScheduled`), and both `settled` arms require it. Verified live afterwards: restarting the OnDelete StatefulSet holds the watch for the full two minutes and closes with "no answer" instead of a green lie — the third state, where there used to be a confident second one.

### The peek fired a managed restart the page would have asked about

`delivery-intercept.tsx` opens its dialog on `intercept !== null || critical`. The peek asked only `criticalActive`, so on an ordinary cluster an Argo-managed StatefulSet restarted from the peek fired straight through while the page for the same object asked "Argo CD will undo this Restart" first. The dialog was already rendering the intercept text; only the decision to open it ignored it. The rule is one exported function now, `restartNeedsAsking`, with a test.

### The ScrollArea guard I added in #211 accepted `h-full`

The exact percentage height the guard exists to forbid. `ownHeight` was `/(^|\s)(h-|max-h-|size-)/`, so `h-full` and `size-full` passed — it caught only the one spelling that happened to be in What's New. It also reported `className={cn("h-64", x)}` as a failure, because it could not read classes out of a call. Both fixed, both covered, and the guard is sabotage-checked against `h-full` specifically.

### Two peek rows were hand-built English

`` `${set.replicas.ready} of ${set.replicas.desired}` `` on the StatefulSet peek and the same shape on the Job peek — concatenated, so no scanner finds them, which is the pattern CLAUDE.md names. The catalogue key was already there and already Russian, used twice in the same file.

### Found and not fixed

Two more hand-built English strings of the same shape, outside the peek and outside what #211/#212 touched: `detail-tab.ts:134` ("N of M failing") and `node-pool.ts:220` ("N of M spot"). Both need a new catalogue string rather than an existing key, so they are not one-liners and I have left them.

The Helm list still drops any secret it cannot decode and reports the survivors as the whole truth, and `get_helm_release_detail` still turns a decode failure into "Release not found", which the frontend renders as "may be gone". The `config` field is fixed; the mechanism is not, and saying "1 release, 2 could not be read" means a shape change through the IPC.


Dudude-bit:
Right, and a hole I made myself. The guard before #214 rejected `min-h-64` — `/(^|\s)(h-|max-h-|size-)/` never matched it, because the `h-` there follows a dash rather than whitespace. I widened the prefix list to carve out `min-h-0`, which never needed carving out since a minimum was never a height here, and that opened a bypass for the same bug in a different spelling.

Fixed in #219 (`b112c50`): `min-h` is out of the list, `min-h-64` and `min-h-[20rem]` are covered, and pointing What's New at `<ScrollArea className="min-h-64">` now fails the guard by name.

PR #221, opened 2026-09-19
fix: what the review confirmed on the list work

Dudude-bit:
The 21 findings an adversarial review confirmed on #220's list work — which I merged before that review finished, by cutting the Flatpak branch off it. That was my mistake, not the tool's.

### The drag did not follow the pointer (high)

The worst of it. TanStack's handler commits pixel deltas; these tables are laid out in shares of their own width, and the dragged column sits in its own denominator — so the rendered edge moved by a fraction of the travel, and the fraction shrank the longer the drag went on. On the Pods list, 200px of drag moved the grip about 100px. **The last column's grip could not move at all**: its right edge is the table's.

A drag now moves width from one column to the next instead of adding it. The total is constant, so a share moves one-for-one with the pointer. The last column has no grip — nothing to its right to take from. `columnResizingFeature` is out of the feature set again, with the reason written where the next person will look.

Three sabotages fail the new tests: adding instead of moving, a grip on the last column, and dropping the write on release.

### `ResourceName` stopped clipping itself (high)

I replaced one `truncate` with a flex row and put nothing back. A tail that cannot shrink paints outside its box — over the copy mark in a list cell, over the next column, into a scrollbar in the peek. The outer span carries `overflow-hidden` now, and columns have a `MIN_COLUMN_SIZE` of 56 so a column cannot be dragged narrower than its own header word.

### Two of my tests could not fail (high, ×2)

Verified by the reviewer with actual mutation, which I should have run myself and did not:

- `carried-search.test.ts` only exercised the pure function. Deleting the sidebar's call to it — the whole feature — left all 26 tests green. It asserts the rendered `href` now, and fails on exactly that deletion.
- The widths store test passed under a merging `set`, and the reset path depends absolutely on replace. It now writes a key and then drops it, and fails when `set` merges.

### The rest

- A drag wrote `localStorage` every frame through a persisted store. It writes once, on release; the frames go to local state.
- Computing the final width *inside* a React state updater meant a drag released in the same tick as its last move lost it. Found by the new test, not by reasoning.
- The grip straddled the column edge, putting 4.5px past the table and giving every list a few pixels of horizontal scroll it never had.
- `?q=` was carried from any page, including Traefik where `q` is a hostname filter — so every kind row there pointed at `/pods?q=shop.example.com`. Gated at both ends now, against the registry's own list routes.
- Double-click reset persisted `{}` instead of forgetting the table.
- Two CRDs sharing a plural shared one width key; `certificates` is both cert-manager's and Knative's. A CRD list files its widths under plural + group.
- The peek's copy mark moved out of `SheetTitle`: Radix points the dialog's `aria-labelledby` there, so a button inside made the panel announce itself as "name Copy name: name".
- The non-routable `ResourceRef` branch gained the `aria-label` the link branch already had, since the name is drawn as two boxes.

### Also here: the Flatpak workflow's first run failed

It checked out the released tag to find `packaging/`, and the recipe only exists on main — so a tag cut before it has no such file. It checks out the workflow's own branch now; only the AppImage comes from the tag.

**Left alone:** dragging a name selection still copies a newline between stem and tail, because flex items are blockified. Fixing that means giving up the layout that keeps the tail visible, which is the point of the change.


Dudude-bit:
Verified live after the rewrite, and it found one more the review had listed and I had not closed.

The drag is 1:1 now — probed it rather than eyeballing: `own` went 320 → 394.4 for 60px of travel, which against a 1438 total in a 1160px table is exactly 60 rendered pixels. Before, 200px of drag moved the edge about 100.

But the first live drag squeezed the neighbour until "Состояние" painted straight over "CPU" — finding #6, which I had answered only by raising the minimum and never by clipping. `TableHead` has `overflow-hidden` now, so a narrowed header is cut at its own boundary instead of over the next one, and the floor is 80 rather than 56. There is a test for the clamp as well as for the move.

PR #222, opened 2026-09-19
fix(connections): a refused list is not an empty one, and not a missing object

Dudude-bit:
Reopens the work of #162, which GitHub closed on its own when its base branch (#161) was deleted on merge. Nothing about it was reviewed and rejected.

The fix is not in main: `Snapshot::of` still does `pods: pods?.items` for pods, services and ingresses, so one refused list kills the whole neighbourhood answer, and an object whose list could not be read is reported as not found. #161 fixed the same shape for claims only.

Needs a rebase on main and carries conflicts — filed so it is not forgotten, not to be merged as it stands.

### DIFF
--- a/src-tauri/src/commands/connections.rs
+++ b/src-tauri/src/commands/connections.rs
@@ -704,7 +704,7 @@ fn uses_from_spec(
     ns: &str,
     subject: &ObjectRef,
     spec: &PodSpec,
-    claims: &[PersistentVolumeClaim],
+    claims: &Read<PersistentVolumeClaim>,
     out: &mut Neighbourhood,
 ) {
     let mut targets: Vec<(String, String)> = Vec::new();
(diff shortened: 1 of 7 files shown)


Dudude-bit:
Ran the review workflow over the branch (4 dimensions, each finding put to independent refutation). What survived is fixed here, and I confirmed the result on a kind cluster under a RBAC context that gets a 403 on Services.

**The refusal it removes from one page, it adds to two others.** `Snapshot::of` stops aborting on a refused pods/services/ingresses list, and the accessors hand those back as empty slices. On the Service page that made `note_reach` say "No pod carries app=shop" in red about pods nobody was allowed to read; on the Ingress page a refused Services list made every backend "a backend that was never created". That is the exact defect this PR exists to remove, moved onto two pages it did not cover. Neither `service_connections` nor `ingress_connections` filled `not_looked_at` either, so on those pages the refusal was invisible as well — an empty `not_looked_at` is the wire contract for "every kind was read", and the frontend guard had nothing to fire on.

**A refused list must keep the words the rest of the app matches on.** `read()` turned every failure into `kube::Error`'s own Display, so a 401 lost the `CREDENTIALS_EXPIRED:` prefix the session-expiry detector matches on, a timed-out read lost `READ_DEADLINE:`, and an API failure printed kube 4's `Status` struct at the reader instead of the cluster's sentence. It goes through the app's `Error` now, which is where both markers are put on. And `found()` passed an empty namespace to `not_found`, so the panel said an object is not there without saying where it looked — a different claim on a cluster where the same name lives in two namespaces.

**Three rules nothing was holding.** The `not_looked_at` test checked that a kind name appears and nothing else, so any kind-to-version pairing passed and a list that answered could have been named beside the refused one. `chainSilence`'s guard was tested for a Pod subject alone, so narrowing it to Pods left a Deployment page saying "No Service in this namespace" about Services nobody read. And the accessors' documented rule — a refusal comes back as an empty slice, which is exactly why the verdict sites must read the `Result` and not the slice — had no test at all.

The merge also restores what main had fixed in the live harness and my resolution reverted: `shell-demo`, which exists and mounts a claim, and the guard that stops the loop passing with zero iterations.

PR #223, opened 2026-09-19
feat(share): an investigation as one file, and the targets it may be published to

Dudude-bit:
Reopens the work of #153, which GitHub closed on its own when its base branch (`feat/outcome`) was deleted on merge. It was never reviewed and rejected.

Two commits of real work, about 2500 lines: an investigation exported as one file with no Secret in it (`report.ts`, `usePodReport`, `ShareDialog`), and publishing targets with a key that stays in the config (`sharing.rs`, `SharingSettings`). None of it is in main.

The branch is 56 commits behind main and carries the whole old stack of merges underneath, so the diff here is not reviewable and taking it wholesale would revert a month of main. It needs its two commits replanted on a fresh branch. Filed so it is not forgotten.

## Summary by CodeRabbit

- **Новые возможности**
  - Добавлена публикация отчётов расследований в настроенные цели.
  - Отчёты можно просматривать, сохранять как HTML и копировать ссылку после публикации.
  - В настройках появился раздел управления целями: создание, редактирование, удаление, проверка и импорт ключа Postplan.
  - Для публичных целей отображается предупреждение и требуется подтверждение публикации.
  - Добавлены английская и русская локализации новых функций.

- **Улучшения**
  - Диагностика Pod и формирование отчётов стали более структурированными.
  - Отчёты поддерживают светлую и тёмную темы, мобильный просмотр и безопасное отображение данных.
  - Улучшена обработка сетевых связей, логов и непрочитанных данных.


Dudude-bit:
Rebuilt and reviewed.

**Rebuilt, because it could not be read.** This branch was the tip of an old stack: eleven features had been merged into it that have since landed on main separately, so git saw two unrelated histories for the same files and offered 104 conflicts, half of them add/add on whole files. Worse, the branch carried *older* copies of code that has been fixed since — its `useChain` predates the cross-namespace address fix, its `MostLikelyPanel` predates the custom search engine — so resolving in its favour would have silently reverted both. Its own two commits are replayed on current main instead, and where the branch pulled backwards I took main's version and re-applied the intent. The hook extraction is done from main's `useChain`, not the branch's.

**The one that matters.** The report wrote log lines raw. `redact()` is documented in this repo as applied to everything the app hands to somebody else, and the file that ends up at a public URL was the one path that skipped it — while its own footer said "No Secret value is ever written into this file." A crash-looping pod whose framework echoes its resolved config prints `dsn=postgres://app:hunter2@db:5432/app`. It happens in the builder now, not at the call site: every caller of `renderReport` is handing the result to somebody.

**Five things the file stated that it did not know.** A refused or still-running connections read reached it as an empty chain and printed "Nothing here." with nothing in "Not read" — the opposite answer, at the moment the page itself says it could not read what connects. The chain listed edges only, so a Service that publishes no endpoint arrived as an ordinary working hop. The status came from `ready`, which is not the status. Node silence never reached the file, so a stale state was written as the state now. And the hedge that says this app was never watching this cluster was decided on a journal that is global.

**And the rest:** a targets list the app could not read was drawn as "you have not added one yet", in the dialog and in Settings both; the capture stamp was minted inside the memo, so every watch tick unticked the public-target acknowledgement and erased the published link; `publicUrl` fell back to the target's own base URL, so a publish with no link handed the reader the service root and called it the report; the reserved red went to any private target hosted on a name containing "prod"; the file declared `lang="en"` whatever it was written in; the character count had one Russian form for every number; `readyToPublish` had no callers while the dialog spelled the rule twice; the report was built before the app knew its own version; and turning «Most likely» off stopped the panel but not the reading behind it — the hook extraction had given the report its own copy of the read.

The test that claimed no Secret can reach the file pushed a field `renderReport` never reads, so it passed against any builder at all; it now asserts where it can fail. `usePodReport` had no test file — it has seven, each of which fails when its rule is deleted.

Gates: tsc, lint, 3093 frontend tests, 726 Rust tests, clippy, both formatters, comment budget 9%.

